PR #1, opened 2023-12-20
Extend `Synchronize()`

friendlymatthew:
Because we want to introduce other format types, first we should identify the JSONL specific logic and try to extend them. 

# What this PR contains:
1) `DetermineDataHandler`: a method that checks the suffix of the filepath to identify the format. 

> *Is `strings.HasSuffix` enough? Or should there be more robust measures like content sniffing?*

<br />

2) Since `Serialize()` contains JSONL specific logic + `handleObject`, I wrote an interface `DataHandler` that implements `Synchronize`

> This is the header: `func (j JSONLHandler) Synchronize(f *IndexFile) error {`. 
> However, it would be nicer to have  `func (j JSONLHandler) Synchronize() error {` where

```go
type JSONLHandler struct{indexFile *IndexFile}
```

> But this would require a lot of refactoring, and I wanted to get your thoughts first. What I currently have fits nicer with the existing flow.

- Refactor tests

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -31,14 +31,8 @@ func main() {
 		panic(err)
 	}
 
-	// Determine file type, assign correct handler
-	handler, err := appendable.DetermineDataHandler(args[0])
-	if err != nil {
-		panic(err)
-	}
-
 	// Open the index file
-	indexFile, err := appendable.NewIndexFile(file, handler)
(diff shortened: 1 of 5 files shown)


friendlymatthew:
> I think we should actually require the user to choose. In the cli we can accept a `-jsonl data.jsonl` flag to point at the data file.

Sounds good. So I removed the `DetermineDataHandler` function from this PR, and I can implement this flag feature in a separate issue.

PR #6, opened 2023-12-26
support CSV

### LINKED ISSUE #4: Add csv support
We'd like to support CSV data files. This will be very similar to json with one minor nuance that the field types are a bit complicated with CSV. We will likely need to document our choices around how fields are parsed/which fields are treated as strings carefully.

https://datatracker.ietf.org/doc/html/rfc4180

friendlymatthew:
fixes #4 

# What this PR does

- [x]  Adds the `--jsonl` or `-csv` flags to the CLI. 
- [x]  `NewIndexFile` and extends `Synchronize` for the `csv` datahandler
- [x]  Refactor `ReadIndexFile` to handle various formats
- [x]  Add tests to check byte-identical, checksums between `csv` Index file <---> `csv` readIndexFile 
- [ ]  Add tests to check byte-identical between `csv` Index file <---> `jsonl` Index file


kevmo314:
This is a pretty good start. One idea for a test would be to generate the index file and check if it's byte-identical to the index file created by the jsonl version.

Since the taxicab data is relatively primitive types, this should be not too crazy to do.

friendlymatthew:
> This is a pretty good start. One idea for a test would be to generate the index file and check if it's byte-identical to the index file created by the jsonl version.
> 
> Since the taxicab data is relatively primitive types, this should be not too crazy to do.

Hey Kevin, 

Sorry about the AWOLness, I've been pretty sick with the flu for the past week. I'm slowly feeling better and will have this feature up soon.

Happy new year :)

kevmo314:
No worries, it's the holidays anyways so everything is on higher latency. I also got sick with covid + a cold a two weeks ago so I know how it is, it seems like everyone has been getting sick :)

Happy new year!

pkg/appendable/io.go
```diff
@@ -109,28 +121,49 @@ func ReadIndexFile(r io.Reader, data DataHandler) (*IndexFile, error) {
...
+			start = f.EndByteOffsets[0]
+			startIndex = 1
+		}
+		for i := startIndex; i < int(ifh.DataCount); i++ {
+
+			fmt.Printf("Current start idx: %d, position: %d\n", i, start)
+
+			if _, isCsv := data.(CSVHandler); isCsv {
+				if i > 1 {
+					start -= 1
```
friendlymatthew, line 152: this is a hotfix solution that worked.

The reason was the start position was +1 from the ideal start byte offset

pkg/appendable/io.go
```diff
@@ -145,10 +178,68 @@ func ReadIndexFile(r io.Reader, data DataHandler) (*IndexFile, error) {
 		if _, err := data.Seek(int64(f.EndByteOffsets[len(f.EndByteOffsets)-1]), io.SeekStart); err != nil {
 			return nil, fmt.Errorf("failed to seek data file: %w", err)
 		}
+
+		if _, isCSV := data.(CSVHandler); isCSV {
+			f.EndByteOffsets = f.EndByteOffsets[1:]
+			f.Checksums = f.Checksums[1:]
+		}
```
friendlymatthew, line 185: we trim the head off for both `f.EndByteOffsets` and `f.Checksums` because the first element are always headers. 

*Why do we need to store them in the first place?*
The start byte offset must start at the end of the header. So this means we need to store the header information in the byte offset.

We store the header checksum to make the relation between `EndByteOffset` and `Checksums` to be bijective

pkg/appendable/io.go
```diff
@@ -254,11 +372,29 @@ func (f *IndexFile) Serialize(w io.Writer) error {
...
+
+			if isCSVHandler {
+				if atr, btr := fieldRankCsvField(at), fieldRank(bt); atr != btr {
+					return atr < btr
+				}
+			} else {
+				if atr, btr := fieldRank(at), fieldRank(bt); atr != btr {
+					return atr < btr
+				}
 			}
```
kevmo314, line 396: Instead of setting the boolean above it would be better to switch on the data handler and rank independently. I think it will be too hard to try to maintain a unified ranking.

```suggestion
		switch f.data.(type) {
		case CSVHandler:
			if atr, btr := fieldRankCsvField(at), fieldRankCsvField(bt); atr != btr {
				return atr < btr
			}
			// TODO: compare at, bt knowing that they are the same type.
		case JSONLHandler:
			if atr, btr := fieldRank(at), fieldRank(bt); atr != btr {
				return atr < btr
			}
			switch at.(type) {
			case nil:
				return false
			case bool:
				return !at.(bool) && bt.(bool)
			case int, int8, int16, int32, int64, float32, float64:
				return at.(float64) < bt.(float64)
			case string:
				return strings.Compare(at.(string), bt.(string)) < 0
			default:
				panic("unknown type")
			}
		default:
			panic("unknown handler")
		}
```

This ranking is probably better to put on `DataHandler` but that can be done in a separate PR.

pkg/appendable/io.go
kevmo314: Interesting, I hadn't considered this. There are some implicit assumptions elsewhere in the code that the start byte offset for the first element is always zero, which isn't true in this case.

I think this won't be a problem when we migrate to the B+ tree structure, but just keep that in mind.

kevmo314 (approved):
Nice work on the tests, makes it super easy to figure out.

pkg/appendable/io.go
friendlymatthew: Yeah that was lazy on my part.

PR #31, opened 2024-01-08
Protocol and Index_File

friendlymatthew:
*Still a draft, I plan on committing frequently for transparency*

The first step was to translate `protocol`.

It should be fairly similar, with the only difference being a custom `FieldFlags` struct. `set()` uses bitwise operations to set flags corresponding to different `FieldType` values. 

`index_file.rs` should be fairly similar to the Go version, with the only difference being the `Index` struct takes `FieldFlag` as an attribute. 

Also, `IndexKey` is supposed to account for the `any` that you see in the Go version.

### DIFF
--- a/appendable-rs/appendable/Cargo.toml
+++ b/appendable-rs/appendable/Cargo.toml
@@ -8,4 +8,8 @@ edition = "2021"
 [dependencies]
 protocol = { path="../protocol" }
 serde_json = "1.0.111"
-twox-hash = "1.6.3"
+xxhash-rust = { version = "0.8.8", features = ["xxh3"] }
+
+[dev-dependencies]
+tempfile = "3.9.0"
+
--- a/appendable-rs/appendable/Cargo.toml
+++ b/appendable-rs/appendable/Cargo.toml
@@ -6,3 +6,4 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+protocol = { path="../protocol" }
--- a/appendable-rs/appendable/Cargo.toml
+++ b/appendable-rs/appendable/Cargo.toml
@@ -7,3 +7,5 @@ edition = "2021"
 
 [dependencies]
 protocol = { path="../protocol" }
+serde_json = "1.0.111"
+twox-hash = "1.6.3"
--- a/appendable-rs/appendable/Cargo.toml
+++ b/appendable-rs/appendable/Cargo.toml
@@ -7,7 +7,6 @@ edition = "2021"
 
 [dependencies]
 protocol = { path="../protocol" }
-serde_json = "1.0.111"
 xxhash-rust = { version = "0.8.8", features = ["xxh3"] }
 
 [dev-dependencies]
--- a/appendable-rs/appendable/src/handler/jsonl_handler.rs
+++ b/appendable-rs/appendable/src/handler/jsonl_handler.rs
@@ -1,34 +1,25 @@
-use crate::index_file::{Index, IndexFile, IndexKey};
+use crate::index_file::Index;
 use crate::io::DataHandler;
 use serde_json::{Deserializer, Map, Value};
 use std::fs::File;
-use std::io::{BufRead, BufReader, Read, Seek, SeekFrom};
-use twox_hash::xxh3::hash64;
+use std::io::{BufRead, BufReader, Seek, SeekFrom};
+use xxhash_rust::xxh3::Xxh3;
 
 pub struct JSONLHandler {
-    pub file: File,
-    pub buffer: Option<Vec<u8>>,
+    reader: BufReader<File>,
+    xxh3: Xxh3,
 }
 impl JSONLHandler {
     pub fn new(file: File) -> Self {
-        JSONLHandler { file, buffer: None }
-    }
-
-    /// we need to read the entire file to buffer to compute byte slices for checksums
-    pub fn read_file_to_buffer(&mut self) -> Result<(), String> {
-        let mut buffer = Vec::new();
-        self.file
-            .read_to_end(&mut buffer)
-            .map_err(|e| e.to_string())?;
-
-        self.buffer = Some(buffer);
-
-        Ok(())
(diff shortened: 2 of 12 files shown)


appendable-rs/appendable/src/handler/jsonl_handler.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub fn read_file_to_buffer(&mut self) -> Result<(), String> {
+        let mut buffer = Vec::new();
+        self.file
+            .read_to_end(&mut buffer)
+            .map_err(|e| e.to_string())?;
+
+        self.buffer = Some(buffer);
+
+        Ok(())
+    }
```
friendlymatthew, line 27: so instead of reading line by line, we store the entire contents of the file in memory with a buffer.

Originally I wanted to read line by line, but as you see below when we compute checksums, we'll need to compute byte slices

kevmo314: It won't be tractible to store the entire file in memory. For the checksums, if we use xxhash then xxhash can be used incrementally so the entire file doesn't have to be checksummed: https://github.com/Cyan4973/xxHash#example

friendlymatthew: Ok. I found the Rust implementation of `xxHash`. That is: https://github.com/DoumanAsh/xxhash-rust?tab=readme-ov-file#example

I'll refactor to use this

appendable-rs/appendable/src/tests/jsonl_index_file.rs
```diff
@@ -0,0 +1,39 @@
...
+            temp_file,
+            r#"{{"name": "matteo", "id": 2, "alpha": ["a", "b", "c"]}}"#
+        )?;
+        writeln!(
+            temp_file,
+            r#"{{"name": "kevin", "id": 1, "alpha": ["x", "y", "z"]}}"#
+        )?;
+
+        // Persist the file and return the File handle
+        let file = temp_file.persist(Path::new("mock_data.jsonl"))?;
```
kevmo314, line 24: Is it possible to do this in memory without having to write a file?

appendable-rs/appendable/src/handler/jsonl_handler.rs
```diff
@@ -0,0 +1,71 @@
...
+use std::fs::File;
+use std::io::{BufRead, BufReader, Seek, SeekFrom};
+use xxhash_rust::xxh3::Xxh3;
+
+pub struct JSONLHandler {
+    reader: BufReader<File>,
+    xxh3: Xxh3,
+}
+impl JSONLHandler {
+    pub fn new(file: File) -> Self {
```
kevmo314, line 13: Can we use a `Cursor` instead of `File` here to also allow non-filesystem-backed implementations? https://doc.rust-lang.org/std/io/struct.Cursor.html

appendable-rs/appendable/Cargo.toml
```diff
@@ -6,3 +6,10 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+protocol = { path="../protocol" }
+serde_json = "1.0.111"
```
kevmo314, line 10: This isn't necessary anymore, right?

kevmo314 (commented):
Direction looks good so far.

appendable-rs/appendable/src/tests/jsonl_index_file.rs
friendlymatthew: Yes this makes sense. However there was a reason why I wanted to persist the file right now.

That is to feed in the `mock_data.jsonl` to our existing Go engine. The goal is to add detailed print statements upon index file creation and do a deep comparison between our Rust and Go versions.

Would it be alright to keep this, create another version that's in memory, then when the Rust version is mature we can remove the above code?

appendable-rs/appendable/src/handler/jsonl_handler.rs
friendlymatthew: Oh awesome -- most definitely!

appendable-rs/appendable/Cargo.toml
friendlymatthew: So I'm still debating whether or not to include `serde_json`. My current work uses it to parse the `json` object and it allows me to cleanly pattern match like so:

```rust
let json: Value = serde_json::from_str(json_line).map_err(|e| e.to_string())?;

    let mut bytes_processed = 0;

    match json {
        Value::Object(obj) => {
            for (k, v) in obj {
                let field_offset = data_offset + (k.len() as u64) + 1; // account for : sym

                path.push(k.clone());
                let name = path.join(".");

                match v {
                    Value::Object(_) => {
                        let nested_json = serde_json::to_string(&v).map_err(|e| e.to_string())?;

                        let nested_bytes = handle_json_object(
                            &nested_json,
                            index_file,
                            path,
                            data_index,
                            field_offset,
                        )?;

                        bytes_processed += nested_bytes
                    }
                    Value::String(s) => {}
                    Value::Number(n) => {}
                    Value::Bool(b) => {}
                    Value::Null => {}
                    _ => return Err(format!("Unexpected type: {}", v)),
                }
            }
        }
        _ => {
            return Err("Expected JSON object".to_string());
        }
    }
```

But as you can see we might not be accurate with our bytes calculation. So I wanted to experiment with the `serde_json` dependency, and if it doesn't mix well, I'd attempt at writing a custom tokenizer

kevmo314: Up to you, but I suspect since we need byte-precise parsing, it might be simpler to write a custom tokenizer. JSON is not too complicated. I'm surprised Rust doesn't have one in the standard library, actually.

appendable-rs/appendable/src/tests/jsonl_index_file.rs
kevmo314: Yeah that's fine

friendlymatthew:
Going to close this after today's sync.

RIP 🦀

Issue #33, opened 2024-01-12
jest testing


friendlymatthew:
Add testing environment, write out tests

friendlymatthew:
Closed via #34, but will add more tests

PR #34, opened 2024-01-12
Add Jest

friendlymatthew:
Solves #33 

## What this PR contains

- [x] A jest test framework built and configured with Typescript.

- [x] A `database.test.ts` that tests a simple query.
     - mocks index file headers, index records, and data records
     - mocks a JSON string from `.get()` in `data-file.ts`
     - builds query

## How to run tests
```sh
npm test
```

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -2,207 +2,210 @@ import { DataFile } from "./data-file";
 import { IndexFile, VersionedIndexFile } from "./index-file";
 
 type Schema = {
-  [key: string]: {};
+	[key: string]: {};
 };
 
-type WhereNode<T extends Schema> = {
-  operation: "<" | "<=" | "==" | ">=" | ">";
-  key: keyof T;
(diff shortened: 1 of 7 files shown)


src/database.ts
kevmo314, line 1: Are you using prettier to format this?

kevmo314 (approved):
Nice thanks, just one question on the formatter.

src/database.ts
friendlymatthew: Yes

friendlymatthew: Should I use a different formatter? Happy to reformat :)

kevmo314: Nope, was just curious since there were some changes. Prettier is fine. We should probably pin the version in dev dependencies but that's beyond this PR

friendlymatthew: > We should probably pin the version in dev dependencies but that's beyond this PR

Noted

PR #35, opened 2024-01-12
btreejs

### LINKED ISSUE #19: Add a js btree reader implementation

friendlymatthew:
fixes #19 

> [!WARNING]
> This is a rough draft, it is untested. Figured I would commit early to verify that this is the correct direction.

## What this PR contains
Implements the searching algorithm for a bptree.

## Next steps
- [ ] add descriptive comments
- [x] add a testing environment
     - Use https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/bytes/bytes_test.go test cases for `compareBytes()`
     - Test binary search
     - Ensure offset calculations are correct


src/btree.ts
```diff
@@ -0,0 +1,135 @@
+import { BPTreeNode, MemoryPointer, compareBytes } from "./node";
+
+// taken from `buffer.go`
+interface ReadWriteSeekTruncater {
+	write(buffer: Uint8Array): Promise<number>;
+	seek(offset: number, whence: "start" | "current" | "end"): Promise<number>;
+	read(buffer: Uint8Array, offset: number): Promise<number>;
```
kevmo314, line 7: The go pattern of writing to a buffer is a bit of an antipattern in javascript. Since the http request allocates a buffer anyways, writing to a buffer will result in a copy which defeats the benefit that passing the output buffer offers.

In other words, the read function is really just a `read(offset: number): Promise<Uint8Array>`, which is what the RangeResolver provides: https://github.com/kevmo314/appendable/blob/main/src/resolver.ts#L16-L23

src/node.ts
```diff
@@ -0,0 +1,184 @@
+class ReferencedValue {
```
kevmo314, line 1: `ReferencedValue` and `MemoryPointer` can be type definitions instead which is more idiomatic for typescript:

```
type ReferencedValue = {
  dataPointer: MemoryPointer;
  value: Buffer;
};

type MemoryPointer = {
  offset: number;
  length: number;
};
```

src/node.ts
```diff
@@ -0,0 +1,184 @@
...
+		this.keys = keys;
+	}
+
+	addPointer(pointer: MemoryPointer) {
+		this.pointers.push(pointer);
+	}
+
+	addKey(key: ReferencedValue) {
+		this.keys.push(key);
+	}
```
kevmo314, line 46: I don't think these are necessary because they're related to writing?

src/node.ts
```diff
@@ -0,0 +1,184 @@
...
+		}
+
+		return [lo, false];
+	}
+}
+
+// https://pkg.go.dev/internal/bytealg#Compare
+export function compareBytes(
+	a: Uint8Array | null,
+	b: Uint8Array | null
```
kevmo314, line 152: It's fine to take these as non-nullable which will help simplify the remainder of the algorithm.

src/node.ts
```diff
@@ -0,0 +1,184 @@
...
+
+				m += 8;
+			}
+		} catch (error) {
+			return 0;
+		}
+		return m;
+	}
+
+	async bsearch(key: Uint8Array): Promise<[number, boolean]> {
```
kevmo314, line 127:
```suggestion
	async bsearch(key: Uint8Array): Promise<number> {
```

In js it's more idiomatic to return the negative insertion index like how Java does it: https://stackoverflow.com/questions/13734694/array-binarysearch-returning-negative-numbers

I'll leave comments below on how to adjust the return value

src/node.ts
```diff
@@ -0,0 +1,184 @@
...
+		let lo = 0;
+		let hi = this.keys.length - 1;
+
+		while (lo <= hi) {
+			const mid = (lo + hi) / 2;
+			const cmp = compareBytes(key, this.keys[mid].value);
+
+			switch (cmp) {
+				case 0:
+					return [mid, true];
```
kevmo314, line 137:
```suggestion
					return mid;
```

Return the index if it was found exactly.

src/node.ts
```diff
@@ -0,0 +1,184 @@
...
+				case -1:
+					hi = mid - 1;
+				case 1:
+					lo = mid + 1;
+			}
+		}
+
+		return [lo, false];
```
kevmo314, line 145:
```suggestion
		return ~lo;
```

Otherwise, return the negative index minus one, which is equivalent to bitwise not (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_NOT).

To use the value in the caller, we currently have something like

```
		const [i, found] = await n.bsearch(key);
		if (found) {
                  // i is the location of key
		}
```

This will instead be

```
		const i = await n.bsearch(key);

		if (i >= 0) {
			// i is the location of key
		} else {
			// ~i is where key should be inserted if it did exist
		}
```

kevmo314 (changes requested):
This is on the right track, don't get dissuaded by the number of comments :)

The main thing I notice is that some of the patterns are more idiomatic in golang than typescript. These are better converted to typescript paradigms which is what my comments mostly remark on. Besides more idiomatic code, it will also be a good way to get exposure to the nuances of each language to follow the patterns in that language. 

Looking forward to the tests too, but for the sake of simplicity I'm also fine with addressing these comments, merging this code, and adding the tests later.

src/node.ts
friendlymatthew: Yeah - my bad.

src/node.ts
friendlymatthew: Dang the ~ is really cool. Thanks!

src/node.ts
friendlymatthew: Sounds good -- I will omit the `nil` cases from the `compare` go tests

https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/bytes/compare_test.go;l=34-39

friendlymatthew:
@kevmo314 

All comments have been addressed, and I've added the necessary tests for binary search and compareBytes.

I'm currently working on testing `readFrom()`. Since it's rather a length PR, I'd prefer if we can merge this in, and I'll post another PR with more tests. 

Hope that is alright.

src/node.ts
```diff
@@ -0,0 +1,190 @@
...
+		this.dataHandler = dataHandler;
+		this.pointers = pointers;
+		this.keys = keys;
+	}
+
+	leaf(): boolean {
+		return this.pointers.length === this.keys.length;
+	}
+
+	async readFrom(): Promise<number> {
```
kevmo314, line 25: This function is probably better off `static`. In practice, the `readFrom()` pattern is also a golang pattern, so something like

```
static async fromMemoryPointer(mp: MemoryPointer, resolver: Resolver): Promise<BPTreeNode> {
```

It's fine to leave for now and address in the next PR.

kevmo314 (approved):

PR #41, opened 2024-01-16
Append index for CSV

### LINKED ISSUE #39: Fix `csv` test case: should append data range for given Index
Relates to:
https://github.com/kevmo314/appendable/blob/165a9ef6934b8b1df941f3d3637ccaa3bdb57be3/pkg/appendable/index_file_csv_test.go#L52-L105

## The issue:
This test case should append an IndexRecord when introduced with another data record. 

It creates a new index file from data: `test\ntest1\n`. Which creates an `Index` with one data range. Then, we call `ReadIndexFile` with data that has an extra data field: `test\ntest1\ntest2`. 

The problem is `ReadIndexFile` calls `Synchronize` inside. `Synchronize` treats the first line as the header, which is ideal when we call `Synchronize` from `NewIndexFile`, but not ideal when we call `Synchronize` from `ReadIndexFile`. 

So the output looks like: 

```sh
=== NewIndexFile from `test\ntest1\n` ===
Header 0 - StartOffset: 0, EndOffset: 5, Checksum: 5754696928334414137
Line 1 - StartOffset: 5, EndOffset: 11, Checksum: 833051405302831824
Field 'test' - Offset: 5, Length: 5, Value: test1, Type: 1

checksums look like: [5754696928334414137 833051405302831824]



=== ReadIndexFile ===

Fields read at offset 5: [test1]
Current start idx: 1, position: 5
string: test1 and bytes look like: [116 101 115 116 49]
endbyteoffsets: [5 11]

Header 0 - StartOffset: 11, EndOffset: 17, Checksum: 7330617272339924708
checksums = [7330617272339924708]
```

friendlymatthew:
Fixes #39 

## The problem
To append a new index file, `ReadIndexFile` is called. `ReadIndexFile` proceeds to build the `IndexFile` from the data generated by `NewIndexFile`. To append new data that exists in `ReadIndexFile` but not in `NewIndexFile`, we call `Synchronize`. The problem is for CSV files, we need to pass in the headers from `ReadIndexFile`.

## The solution

1. Extract headers from data passed in from `ReadIndexFile`. 
> We could have spliced the data from 0 -> `f.EndByteOffsets[0]`. But since we want to support a schemaless design, if the headers change, we want to catch for this. So we need to use `csvReader` to parse the header.

2. Pass headers to `Synchronize`
> `Synchronize` gets called in two different places:
     
     1. Index file creation, here we can strip the head line and treat it as our header
     2. Read index file, which needs explicit header data passed

## Next steps
- [ ] write more tests, the existing append index file test passes
- [ ] clean up, remove debug comments


pkg/appendable/csv_handler.go
```diff
@@ -17,12 +17,18 @@ type CSVHandler struct {
 	io.ReadSeeker
 }
 
-func (c CSVHandler) Synchronize(f *IndexFile) error {
-
+func (c CSVHandler) Synchronize(f *IndexFile, headerData []string) error {
```
kevmo314, line 20: The headers should be able to be determined from the index file itself since they're already parsed.

https://github.com/kevmo314/appendable/blob/165a9ef6934b8b1df941f3d3637ccaa3bdb57be3/pkg/protocol/protocol.go#L49
https://github.com/kevmo314/appendable/blob/165a9ef6934b8b1df941f3d3637ccaa3bdb57be3/pkg/appendable/index_file.go#L16

Basically, when parsing the lines, if the `Indexes` are not present then parse the line as `Indexes` (in other words, read the headers). If the `Indexes` are present, then the line represents a data line. This will allow the handler to not need to take in the headers as an argument.

This will be true in the B+ tree too (although it's called the MetaPage) there.

friendlymatthew: Damn, this is super big brain. Makes things so much nicer. Thanks Kevin!

kevmo314 (approved):

PR #42, opened 2024-01-17
More CSV tests

friendlymatthew:
- [x] null types
- [x] same index but different types
- [x] same index but different types, one type being null
- [x] append with multiple headers
- [ ] append new column (header with data)

### DIFF
--- a/pkg/appendable/index_file_csv_test.go
+++ b/pkg/appendable/index_file_csv_test.go
@@ -165,7 +165,7 @@ func TestAppendDataRowCSV(t *testing.T) {
 	// this test case fails due to the way we measure checksums
 	// since CSV's are column-based appending new columns, cause checksums and endbyteoffsets to be mismatched
 	t.Run("new index", func(t *testing.T) {
-		i, err := NewIndexFile(CSVHandler{ReadSeeker: strings.NewReader(mockCsv)})
+		i, err := NewIndexFile(CSVHandler{ReadSeeker: strings.NewReader("header1\ntest1\n")})
 		if err != nil {
 			t.Fatal(err)
 		}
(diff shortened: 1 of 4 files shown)


pkg/appendable/index_file_csv_test.go
```diff
@@ -122,4 +162,91 @@ func TestAppendDataRowCSV(t *testing.T) {
...
+		}
+
+		buf := &bytes.Buffer{}
+
+		if err := i.Serialize(buf); err != nil {
+			t.Fatal(err)
+		}
+
+		j, err := ReadIndexFile(buf, CSVHandler{ReadSeeker: strings.NewReader("header1,header2\ntest1,test2\n")})
+		if err != nil {
```
friendlymatthew, line 180: mockCsv looks like
```sh
"header1\ntest1\n"
```

which causes every line's checksum + endbyteoffset to be mismatched

pkg/appendable/index_file_csv_test.go
```diff
@@ -122,4 +162,91 @@ func TestAppendDataRowCSV(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		// check that the index file now has the additional data ranges but same number of indices
+		if len(j.Indexes) != 2 {
+			t.Errorf("got len(i.Indexes) = %d, want 2", len(j.Indexes))
+		}
+
+		if j.Indexes[0].FieldType != protocol.FieldTypeNull|protocol.FieldTypeString {
```
kevmo314, line 247: Ooh this is an interesting case. For strings, I guess this makes it impossible to disambiguate between null and empty string? It's fine to treat empty as null, but we should definitely document this choice as I could see a user trying to query for empty.

kevmo314 (approved):
Nice!

pkg/appendable/index_file_csv_test.go
```diff
@@ -122,4 +162,95 @@ func TestAppendDataRowCSV(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		buf := &bytes.Buffer{}
+
+		if err := i.Serialize(buf); err != nil {
+			t.Fatal(err)
+		}
+
+		j, err := ReadIndexFile(buf, CSVHandler{ReadSeeker: strings.NewReader("header1,header2\ntest1,test2\n")})
```
friendlymatthew, line 179: this comment will probably get lost in the ether, so I'll make sure to document this. But for CSV we'll assume this never happens. 

That is, appending a new column.

PR #45, opened 2024-01-19
Refactor Btree.js

### LINKED ISSUE #37: Refactor JS btree reading logic
- [x] This function is probably better off `static`. In practice, the `readFrom()` pattern is also a golang pattern, so something like

```
static async fromMemoryPointer(mp: MemoryPointer, resolver: Resolver): Promise<BPTreeNode> {
```

is probably more javascript-y.

            
- [x] add test for `read`

friendlymatthew:
Fixes #37 

--
- [x] refactor to read in static method
- [x] removed the need of `rangeResolver` in `BPTreeNode` class
- [x] write e2e tests


src/node.ts
```diff
@@ -22,15 +16,15 @@ export class BPTreeNode {
...
+		mp: MemoryPointer,
+		resolver: RangeResolver
+	): Promise<{ node: BPTreeNode | null; bytesRead: number }> {
 		try {
-			console.log("Fetching initial data...");
-
-			let { data: sizeData } = await this.dataHandler({
-				start: 0,
-				end: 4,
+			let node = new BPTreeNode([], []);
```
kevmo314, line 24:
```suggestion
			const node = new BPTreeNode([], []);
```

For the most part we should try to prefer `const` whenever possible in js

src/node.ts
```diff
@@ -39,39 +33,29 @@ export class BPTreeNode {
 			let leaf = size < 0;
 			let absSize = Math.abs(size);
 
-			console.log(`Size: ${size}, Leaf: ${leaf}`);
-
-			this.pointers = new Array(absSize + (leaf ? 0 : 1))
+			node.pointers = new Array(absSize + (leaf ? 0 : 1))
```
kevmo314, line 36: Since we iterate over the keys down below, there's no need to preallocate the arrays here. These can be removed

src/node.ts
```diff
@@ -80,70 +64,55 @@ export class BPTreeNode {
...
 						end: currentOffset + l,
 					});
 
-					console.log(
-						"key value from buffer: ",
-						Buffer.from(keyValue).toString()
-					);
-					this.keys[idx].value = Buffer.from(keyValue);
-					this.keys[idx].dataPointer.length = l; // directly assign length here
+					node.keys[idx].value = Buffer.from(keyValue);
```
kevmo314, line 82: Then here do something like `node.keys.push({ value: ..., dataPointer: ... });`

kevmo314 (approved):
Looks good, nice work.

friendlymatthew:
Was delayed with this commit, I thought there was a bug that never existed 🤦

Issue #46, opened 2024-01-19
Improve querying experience


friendlymatthew:
Upon someone clicking the query execution button, we can check if it's a valid query.

We call it a valid query iff
- a `where` clause exists
- `operation`, `key`, `value` are the keys to the `where` object
- an optional `orderBy` clause
- if the `orderBy` clause is added, we need the `key` and `direction`.

If these aren't fit, we can disable the button and add helpful statements that instruct how to build proper query :)

PR #52, opened 2024-01-22
Enhance demo

friendlymatthew:
Fixes: #46 

## What this PR contains
- [x] update build script to handle the `-jsonl` flag
- [x] upon `Execute` query being clicked, the query gets validated first. If the query is not valid it will emit a message like this:
Screen Shot 2024-01-21 at 8 48 49 PM
Screen Shot 2024-01-21 at 8 49 05 PM

- [x] when you edit a key value, it synchronizes and updates the other key value so you don't need to edit both the `where` key value and `orderBy` key value. This is nice because the same field we filter by must be used for sorting
- [x] css work so overall demo is more structured

### DIFF
--- a/examples/client/index.html
+++ b/examples/client/index.html
@@ -1,134 +1,316 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/ace/1.4.12/ace.js" type="text/javascript" charset="utf-8"></script>
+	<head>
+		<script
+			src="https://cdnjs.cloudflare.com/ajax/libs/ace/1.4.12/ace.js"
+			type="text/javascript"
+			charset="utf-8"
+		></script>
+		<script src="appendable.min.js"></script>
+		<script>
+			Appendable.init(
+				"green_tripdata_2023-01.jsonl",
+				"green_tripdata_2023-01.jsonl.index"
+			).then(async (db) => {
+				let dbFields = new Set();
+				let fieldTypes = {};
 
-    <script src="appendable.min.js"></script>
-    <script>
-      Appendable.init(
-        "green_tripdata_2023-01.jsonl",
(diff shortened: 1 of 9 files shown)


examples/client/index.html
```diff
@@ -1,134 +1,316 @@
...
-        "orderBy": [
-          {
-            "key": "trip_distance",
-            "direction": "ASC"
-          }
-        ]
-      }, null, 2), -1);
-    </script>
-  </body>
+	<head>
```
kevmo314, line 3: Can you format this with `prettier`?

kevmo314 (approved):

examples/client/index.html
friendlymatthew: So my prettier says `index.html` has been formatted:

```
["INFO" - 10:17:20 PM] File Info:
{
  "ignored": false,
  "inferredParser": "html"
}
["INFO" - 10:17:20 PM] No local configuration (i.e. .prettierrc or .editorconfig) detected, falling back to VS Code configuration
["INFO" - 10:17:20 PM] Prettier Options:
{
  "arrowParens": "always",
  "bracketSpacing": true,
  "endOfLine": "lf",
  "htmlWhitespaceSensitivity": "css",
  "insertPragma": false,
  "singleAttributePerLine": false,
  "bracketSameLine": false,
  "jsxBracketSameLine": false,
  "jsxSingleQuote": false,
  "printWidth": 80,
  "proseWrap": "preserve",
  "quoteProps": "as-needed",
  "requirePragma": false,
  "semi": true,
  "singleQuote": false,
  "tabWidth": 2,
  "trailingComma": "es5",
  "useTabs": true,
  "embeddedLanguageFormatting": "auto",
  "vueIndentScriptAndStyle": false,
  "filepath": "/Users/matthew/Desktop/GARAGE/appendable/examples/client/index.html",
  "parser": "html"
}
["INFO" - 10:17:20 PM] Formatting completed in 41ms.
```

PR #56, opened 2024-01-22
Add `decodeType` to Appendable api

friendlymatthew:
To enforce type validation for the demo's query, we needed a function that can decode the bitmask value into it's respective field rank.
Screen Shot 2024-01-22 at 1 05 44 PM

I figured it would be nice to have this in the API, as other people can directly call `db.decodeType(fieldType)`. 

Wrote tests


src/database.ts
```diff
@@ -203,4 +213,17 @@ export class Database<T extends Schema> {
 			}
 		}
 	}
+
+	// given a bitmask, we decode and return the types
+	decodeType(bitmask: bigint): Set<number> {
```
friendlymatthew, line 218: not a static function as we would want an instance of `database`,  even though this won't vary between various indexfiles.

examples/client/index.html
```diff
@@ -88,8 +88,34 @@
...
-					// assert `where` type
+					const allowedRanks = fieldTypes[whereNode.key];
+
+					let valueRank;
+					let valueType;
+
+					if (whereNode.value === null) {
+						valueRank = 1;
+						valueType = "null";
+					}
```
kevmo314, line 99: This check can be simplified a bit by directly checking the bits:

```suggestion
const FIELD_TYPE_NULL = (1 << 0);

...

					if (whereNode.value === null) {
					  if ((field.fieldType & FIELD_TYPE_NULL) > 0) {
					    return <error>;
					  }
					}
```

kevmo314 (commented):
Nice work figuring this out! There's a bit of a simpler way to check the type that maybe we can take advantage of?

examples/client/index.html
```diff
@@ -88,8 +88,36 @@
...
 
-					// todo: find out how to decode bigint field types. i.e. what is a 34n?
-					// assert `where` type
+					const keyType = fieldTypes[whereNode.key];
+
+					const FIELD_TYPE_STRING = 1 << 0;
+					const FIELD_TYPE_NUMBER = 1 << 1;
+					const FIELD_TYPE_BOOLEAN = 1 << 4;
+					const FIELD_TYPE_NULL = 1 << 5;
+
```
friendlymatthew, line 97: Wouldn't it be nice if we could import these constans from the Appendable module?

Or will that just be extra fluff we don't need.

kevmo314: Yeah, that makes sense. We can use a typescript enum, see https://www.typescriptlang.org/docs/handbook/enums.html#computed-and-constant-members

kevmo314:
```ts
export enum FieldType {
  Null = 1 << 0,
  ...
}
```

will let you do `Appendable.FieldType.Null` in the js.

src/index.ts
```diff
@@ -1,22 +1,36 @@
...
+	Appendable: {
+		init: Function;
+		FieldType: typeof FieldType;
+		containsType: typeof containsType;
+	};
+}
+
+declare global {
+	var Appendable: GlobalMap["Appendable"];
 }
```
friendlymatthew, line 31: some type gymnastics

src/index.ts
```diff
@@ -1,22 +1,36 @@
...
+interface GlobalMap {
+	Appendable: {
+		init: Function;
+		FieldType: typeof FieldType;
+		containsType: typeof containsType;
+	};
+}
+
+declare global {
+	var Appendable: GlobalMap["Appendable"];
```
friendlymatthew, line 29: some type gymnastics

src/database.ts
```diff
@@ -21,6 +21,18 @@ export type Query<T extends Schema> = {
...
+	Number = 1 << 1,
+	Boolean = 1 << 4,
+	Null = 1 << 5,
+}
+
+// given a fieldType and the desired type, this function performs a bitwise operation to test membership
+export function containsType(fieldType: bigint, desiredType: FieldType) {
+	return (fieldType & BigInt(desiredType)) !== BigInt(0);
+}
+
```
friendlymatthew, line 35: wondering if this is too extra. 

The reason why I wanted to add this is at first glance it may not be immediate that to check for validity of types we need to perform a bitwise membership test.

src/index.ts
kevmo314: Interesting, I'm surprised this isn't automatic.

kevmo314 (approved):

PR #60, opened 2024-01-23
Switch server

friendlymatthew:
As I was debugging, I was running into a major blocker where after editing the Appendable library, I would rebuild and reserve the demo.

`http-server` must have some underlying caching time, but basically I would have to wait 6 minutes till I saw the following change. Instead of digging into modifying the configuration, I just created a `server.go`.

Now you can just run:
```sh
npm run example
```


kevmo314 (approved):
Haha I love it.

PR #62, opened 2024-01-24
Add logger

friendlymatthew:
Currently doing deep equality index file tests and investigating why CSV start byte offsets are incremented by 1. 

Realized I left some really good print statements that could've been used now, so I factored in a logger. To trigger debug statements, you can simply run: 

```sh
go run cmd/main.go -debug -jsonl examples/workspace/green_tripdata_2023-01.jsonl
```

- [x] refactored tests
- [x] verified index file creation from CLI works

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -21,13 +21,14 @@ func main() {
 
 	flag.Parse()
 
-	var logLevel = new(slog.LevelVar)
+	logLevel := &slog.LevelVar{}
 
 	if debugFlag {
 		logLevel.Set(slog.LevelDebug)
 	}
 
-	var logger = slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: logLevel}))
+	logger := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: logLevel}))
+	slog.SetDefault(logger)
 
 	var totalStart, readStart, writeStart time.Time
 	if showTimings {
@@ -72,7 +73,7 @@ func main() {
 		readStart = time.Now()
 	}
 	// Open the index file
-	indexFile, err := appendable.NewIndexFile(dataHandler, logger)
+	indexFile, err := appendable.NewIndexFile(dataHandler)
 
 	if err != nil {
 		panic(err)
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -4,23 +4,42 @@ import (
 	"bufio"
 	"flag"
 	"fmt"
-	"log"
 	"os"
 	"time"
 
 	"github.com/kevmo314/appendable/pkg/appendable"
+	"go.uber.org/zap"
 )
 
 func main() {
+	var debugFlag, jsonlFlag, csvFlag, showTimings bool
 
-	var jsonlFlag bool
-	var csvFlag bool
-
+	flag.BoolVar(&debugFlag, "debug", false, "Use logger that prints at the debug-level")
 	flag.BoolVar(&jsonlFlag, "jsonl", false, "Use JSONL handler")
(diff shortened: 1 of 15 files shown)


cmd/main.go
```diff
@@ -4,24 +4,43 @@ import (
...
+	var err error
+
+	if debugFlag {
+		logger, err = zap.NewDevelopment()
+
+		if err != nil {
+			panic("cannot initialize zap logger: " + err.Error())
+		}
+	} else {
+		logger, err = zap.NewProduction()
```
friendlymatthew, line 34: prod will only print `info` level logs

cmd/main.go
```diff
@@ -4,24 +4,43 @@ import (
 	"bufio"
 	"flag"
 	"fmt"
-	"log"
 	"os"
 	"time"
 
 	"github.com/kevmo314/appendable/pkg/appendable"
+	"go.uber.org/zap"
```
kevmo314, line 11: Ah, can we use slog? It was added in the latest version of golang https://go.dev/blog/slog, one fewer dependency.

kevmo314 (changes requested):

cmd/main.go
friendlymatthew: Ok

cmd/main.go
```diff
@@ -4,24 +4,31 @@ import (
 	"bufio"
 	"flag"
 	"fmt"
-	"log"
+	"log/slog"
```
friendlymatthew, line 7: retested using cli commands with and without `-debug`

cmd/main.go
```diff
@@ -4,24 +4,31 @@ import (
...
+	flag.BoolVar(&debugFlag, "debug", false, "Use logger that prints at the debug-level")
 	flag.BoolVar(&jsonlFlag, "jsonl", false, "Use JSONL handler")
 	flag.BoolVar(&csvFlag, "csv", false, "Use CSV handler")
-
-	var showTimings bool
 	flag.BoolVar(&showTimings, "t", false, "Show time-related metrics")
 
+	flag.Parse()
+
+	var logLevel = new(slog.LevelVar)
```
kevmo314, line 24:
```suggestion
	logLevel := &slog.LevelVar{}
```

cmd/main.go
```diff
@@ -4,24 +4,31 @@ import (
...
+
+	if debugFlag {
+		logLevel.Set(slog.LevelDebug)
+	}
+
+	var logger = slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: logLevel}))
```
kevmo314, line 30:
```suggestion
	logger := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: logLevel}))
```

cmd/main.go
```diff
@@ -4,24 +4,31 @@ import (
...
+
```
kevmo314, line 31: You can call `slog.SetDefault(logger)` to avoid needing to pass the logger around and instead call `slog.Info()`, etc.

pkg/appendable/index_file_test.go
```diff
@@ -21,10 +23,22 @@ jsonl <---> csv
...
+	originalLogger := slog.Default()
+
+	// Create a logger with Debug on
+	debugLevel := &slog.LevelVar{}
+	debugLevel.Set(slog.LevelDebug)
+	debugLogger := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{
+		Level: debugLevel,
+	}))
+
+	slog.SetDefault(debugLogger)
```
friendlymatthew, line 35: To elevate debug loggers, I'll be using the following logic

kevmo314 (approved):

PR #67, opened 2024-01-26
CSV demo + support null fields

### LINKED ISSUE #65: Support `null` fields
In the `green_tripdata` dataset, all values of the `ehail_fee` attribute are null. When we `Synchronize`, we lose this `Index`. 

As shown here:
```sh
index headers: [VendorID lpep_pickup_datetime lpep_dropoff_datetime store_and_fwd_flag RatecodeID PULocationID DOLocationID passenger_count trip_distance fare_amount extra mta_tax tip_amount tolls_amount improvement_surcharge total_amount payment_type trip_type congestion_surcharge], len: 19
```

We need to support fields even if they're `null`

friendlymatthew:
## What this PR does
- [x] Fixes #65 
> We need to append data even including null fields. This meant modifying the `handleObject` function logic as well as mapping the `null` value to it's correct `protocol.FieldTypeNull` value.
- [x] wrote test to assert claim

- [x] A demo that uses `csv`
> The following screenshot is the demo from `csv` data: 
> Screen Shot 2024-01-26 at 5 17 36 PM

- [x] some helpful scripts for local development setting

## Todo
- [x] perform the same logic for JSONL
- [ ] add jest tests for formatting CSV line
- [x] clean up CSV parsing logic


pkg/appendable/csv_handler.go
```diff
@@ -146,26 +147,37 @@ func (i *IndexFile) handleCSVLine(dec *csv.Reader, headers []string, path []stri
...
+
+			found := false
 			for j := range i.Indexes {
 				if i.Indexes[j].FieldName == name {
 					i.Indexes[j].FieldType |= protocol.FieldTypeNull
+
+					found = true
 				}
 			}
+
```
friendlymatthew, line 171: this logic fixes the #65 bug

PR #74, opened 2024-01-29
Add CSV table

friendlymatthew:
Screen Shot 2024-01-29 at 6 16 18 PM

- [x] responsive and cleanly displays all headers when specified

### DIFF
--- a/examples/client/index.html
+++ b/examples/client/index.html
@@ -54,28 +54,16 @@
 					}
 
 					const query = db.query(queryJson);
-
-					let queryHeaders = [];
-					if (queryJson.select) {
-						queryHeaders = queryJson.select;
-					} else {
-						queryHeaders = Array.from(dbFields);
-					}
-
+					let queryHeaders = queryJson.select ?? Array.from(dbFields);
 					bindQuery(query, queryHeaders);
 				};
 
 				document.getElementById("results").innerHTML = "";
 
 				const queryJson = JSON.parse(editor.getValue());
-				if (queryJson.select) {
-					queryHeaders = queryJson.select;
-				} else {
-					queryHeaders = Array.from(dbFields);
-				}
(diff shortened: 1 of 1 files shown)


examples/client/index.html
```diff
@@ -53,14 +54,29 @@
...
 
 					const query = db.query(queryJson);
-					bindQuery(query);
+
+					let queryHeaders = [];
+					if (queryJson.select) {
+						queryHeaders = queryJson.select;
+					} else {
+						queryHeaders = Array.from(dbFields);
+					}
```
kevmo314, line 63:
```suggestion
					const queryHeaders = queryJson.select ?? Array.from(dbFields);
```

examples/client/index.html
```diff
@@ -53,14 +54,29 @@
...
+
+					bindQuery(query, queryHeaders);
 				};
 
 				document.getElementById("results").innerHTML = "";
 
+				const queryJson = JSON.parse(editor.getValue());
```
kevmo314, line 70: Here too

examples/client/index.html
```diff
@@ -53,14 +54,29 @@
...
+				if (queryJson.select) {
+					queryHeaders = queryJson.select;
+				} else {
+					queryHeaders = Array.from(dbFields);
+				}
+
 				const query = db.query(JSON.parse(editor.getValue()));
 
-				bindQuery(query);
+				bindQuery(query, queryHeaders);
```
kevmo314, line 79: I think it would be reasonable to expose a `query.fields()` function that returns the list of selected fields in the API. That would simplify this a bit.

kevmo314 (approved):

examples/client/index.html
friendlymatthew: I'll address this in a PR that's focused on the API features

PR #77, opened 2024-01-30
Add query validation to library

friendlymatthew:
`query-validation.ts` checks the query structure and style. We add this validation layer inside `db.query()` but before searching starts. This allows us to terminate early iff the validation fails.

The demo shows how you can propagate and display the validation errors.


src/db/query-validation.ts
```diff
@@ -0,0 +1,124 @@
...
+	for (const whereNode of query.where) {
+		if (!["<", "<=", "==", ">=", ">"].includes(whereNode.operation)) {
+			throw new Error("Invalid operation in 'where' clause.");
+		}
+
+		if (typeof whereNode.key !== "string") {
+			throw new Error("'key' in 'where' clause must be a string.");
+		}
+
+		if (!headerNames.has(whereNode.key)) {
```
kevmo314, line 51: I think it's fine to not ache the names and type maps and instead iterate, it makes it easier to reason about.

```suggestion
                const header = headers.find((header) => header.name === whereNode.key);
		if (!header) {
```

src/db/query-validation.ts
```diff
@@ -0,0 +1,124 @@
...
+			throw new Error(
+				`key: ${whereNode.key} in 'where' clause does not exist in dataset.`
+			);
+		}
+
+		if (typeof whereNode.value === "undefined") {
+			throw new Error("'value' in 'where' clause is missing.");
+		}
+
+		const headerType = headerTypeMap[whereNode.key];
```
kevmo314, line 61: Following the above suggestion this would then simplify to

```suggestion
		const headerType = header.fieldType;
```

kevmo314 (approved):
Nice work!

src/db/query-validation.ts
```diff
@@ -0,0 +1,178 @@
...
+ * @param {Header[]} headers - The headers against which to validate the query fields.
+ * @throws {Error} Throws an error if query is invalid.
+ */
+export async function validateQuery<T extends Schema>(
+	query: Query<T>,
+	headers: Header[]
+): Promise<void> {
+	validateWhere(query.where, headers);
+	validateOrderBy(query.orderBy, query.where![0].key as string);
+	validateSelect(query.select, headers);
```
friendlymatthew, line 177: should be good to go! Splitting into helper functions will be useful when we start chaining commands when we query

kevmo314: Very clean very nice.

kevmo314 (approved):

PR #78, opened 2024-01-31
feat: chain queries

friendlymatthew:
## Query Chaining
Query chaining is more fluent and flexible in its query style. 

You can now write queries:
> ```typescript
> const query = await db
>   .where('trip_distance', '>=', 10)
>   .orderBy('trip_distance', 'ASC')
>   .select([
>       'VendorID',
>       'trip_distance',
>       'passenger_count',
>       'fare_amount',
>       'tip_amount',
>       'mta_tax'
>   ])
>   .get();
> ```

It's forkable, so you can write derived queries:
> ```typescript
> const query = await db.where("...");
> const derivedQuery = query.orderBy("...");
>```

This PR also introduces a JS editor to the demo. The editor can now toggle between JSON and JS. Depending on the editor, you can build a query in JSON form or through the chaining of clauses.

Since the demo's gotten a lot more features, I also separated the `.css` and the editor-related logic to its own files.

### DIFF
--- a/src/db/query-builder.ts
+++ b/src/db/query-builder.ts
@@ -1,0 +1,84 @@
+import { Database, OrderBy, Query, Schema, WhereNode } from "./database";
+
+/**
+ * A class for building and executing database queries in a flexible API style.
+ * Allows chaining methods for 'where', 'orderBy', 'select', and 'limit' clauses.
+ */
+export class QueryBuilder<T extends Schema> {
+	private queryObject: Query<T> = {
+		where: [],
+		orderBy: undefined,
+		select: undefined,
+		limit: undefined,
+	};
+
+	/**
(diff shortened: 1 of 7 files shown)


src/db/query-builder.ts
```diff
@@ -0,0 +1,84 @@
...
+
+	/**
+	 * Executes the constructed query and returns an asychronous generator.
+	 * @yields {Promise<IteratorResult<T, any>>} Query result.
+	 */
+	async *get() {
+		for await (const item of this.database.query(this.queryObject)) {
+			yield item;
+		}
+	}
```
kevmo314, line 29:
```suggestion
	get() {
	  return this.database.query(this.queryObject);
	}
```

This can directly return the db object.

kevmo314 (approved):
Nice!

PR #79, opened 2024-01-31
Coverage: Assert query chain behavior

friendlymatthew:
Wrote a testing crate for the new `query-builder.ts`.

Fixed QueryBuilder to return a new instance every time we append another clause. This way each query has its own unique state.


src/tests/query-builder.test.ts
```diff
@@ -0,0 +1,175 @@
...
+		query = q1.toQuery();
+
+		expect(query.where).not.toBeNull();
+		expect(query.where).toEqual([
+			{ key: "fare_amount", operation: "==", value: 1 },
+			{ key: "VendorID", operation: "==", value: 3 },
+		]);
+	});
+
+	it(`test green + red queries`, async () => {
```
friendlymatthew, line 143: an important test case. Create a set of red and green queries from a derived state. Half of them should fail while the other half should pass. But they both originate from the same state. 

This test asserts the uniqueness of a query instance.

src/db/query-builder.ts
```diff
@@ -38,8 +67,9 @@ export class QueryBuilder<T extends Schema> {
 		operation: WhereNode<T>["operation"],
 		value: T[keyof T]
 	): QueryBuilder<T> {
-		this.queryObject.where?.push({ key, operation, value });
-		return this;
+		const currQuery = this.clone();
+		currQuery.queryObject.where?.push({ key, operation, value });
+		return currQuery;
```
kevmo314, line 72: The entire object doesn't necessarily need to be cloned here, just the where field.

```suggestion
                const newQuery = new QueryBuilder(this.queryObject);
                newQuery.where = [...this.queryObject.where, {key, operation, value}];
                return newQuery;
```

Something like this would probably save a bit of memory usage.

kevmo314 (approved):

PR #83, opened 2024-02-05
fix: CI

friendlymatthew:
With the `-i`, we can write the index file to `examples/client/green_tripdata_2023-01.csv.index` removing the need to copy the index file to `examples/client`.

> [!WARNING]
> The `btree` refactor causes the demo to break so I'm a bit cautious about merging this. Ideally we should update the library before merging.


kevmo314 (approved):

kevmo314:
I think it's fine to break the demo since we're relatively unknown, but up to you

friendlymatthew:
Closing due to #118

PR #90, opened 2024-02-06
feat: memory map data file

kevmo314:
This switches the data file to memory mapped access so it doesn't require a copy.

### DIFF
--- a/pkg/handlers/csv.go
+++ b/pkg/handlers/csv.go
@@ -26,7 +26,7 @@ func (c CSVHandler) Format() appendable.Format {
 	return appendable.FormatCSV
 }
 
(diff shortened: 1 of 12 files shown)


pkg/handlers/csv.go
```diff
@@ -57,38 +52,35 @@ func (c CSVHandler) Synchronize(f *appendable.IndexFile, df appendable.DataFile)
 		headers = fieldNames
 	}
 
-	scanner := bufio.NewScanner(df)
-
-	for scanner.Scan() {
-		line := scanner.Bytes()
```
friendlymatthew, line 63: wow this is really smooth

friendlymatthew (approved):

PR #97, opened 2024-02-09
indexfile tsreader

friendlymatthew:
Extremely rough draft


src/index-file.ts
```diff
@@ -30,22 +32,7 @@ export class IndexFile {
 	static async forResolver<T = any>(
 		resolver: RangeResolver
 	): Promise<VersionedIndexFile<T>> {
-		const response = await resolver({ start: 0, end: 0 });
-		const version = new DataView(response.data).getUint8(0);
-		switch (version) {
-			case 1:
```
friendlymatthew, line 36: I took this part out as we read the metadata stored in the `_tree`

src/index-file.ts
```diff
@@ -55,6 +42,57 @@ function decodeFloatingInt16(x: number) {
...
+	constructor(private resolver: RangeResolver) {}
+
+	async tree(): Promise<LinkedMetaPage> {
+		if (this._tree) {
+			return this._tree;
+		}
+		// read the linked meta page
+
+		// we'll have to read the tree file: see multi.ts
+
```
friendlymatthew, line 69: how do we ingest the data from LinkedMetaPage?

src/index-file.ts
```diff
@@ -55,6 +42,57 @@ function decodeFloatingInt16(x: number) {
...
+		const readOffset = dataView.getBigUint64(2);
+
+		return {
+			version: version,
+			format: format,
+			readOffset: readOffset,
+		};
+	}
+}
```
friendlymatthew, line 93: I'm confident in this function. The inner logic is derived from `LinkedMetaPage`'s `metadata()`

src/btree/multi.ts
```diff
@@ -0,0 +1,78 @@
...
+		// reard from 4 to 4 + length for metadata
+		const { data: metadata } = await this.resolver({
+			start: this.offset + 28,
+			end: this.offset + 28 + metadataLength,
+		});
+
+		return metadata;
+	}
+
+	async next(): Promise<LinkedMetaPage | null> {
```
friendlymatthew, line 55: how do we traverse through a linked meta page? 

My thought was we use this when we initially deserialize the tree

src/btree/multi.ts
```diff
@@ -0,0 +1,78 @@
...
+		const { data } = await this.resolver({
+			start: this.offset + 12,
+			end: this.offset + 12 + 8,
+		});
+
+		if (data.byteLength !== 8) {
+			return null;
+		}
+
+		const view = new DataView(data);
```
friendlymatthew, line 65: I used `DataView` as this was frequently used in the previous index file version.

I followed the same pattern, retrieve the ArrayBuffer data from the range resolver, assert the byte length, then feed it into the data view.

src/btree/pagefile.ts
```diff
@@ -0,0 +1,44 @@
+import { RangeResolver } from "../resolver";
+
+const MAX_FREE_PAGE_INDICIES = 512;
+const PAGE_SIZE_BYTES = 4096;
+
+export class PageFile {
```
friendlymatthew, line 6: feel like PageFile reader needs the most work. But I'm just generally stuck on how to approach this class

src/btree/multi.ts
```diff
@@ -0,0 +1,78 @@
...
+		return {
+			offset: pointerOffset,
+			length: lengthOffset,
+		};
+	}
+
+	async metadata(): Promise<ArrayBuffer> {
+		const { data: lengthData } = await this.resolver({
+			start: this.offset + 24,
+			end: this.offset + 28,
```
kevmo314, line 38: It's reasonable to assume that the metadata is the remainder of the page, so the resolver only needs to be called once. That is, `start: this.offset + 24, end: this.offset + 4096 - 1`.

The first four bytes represent the length and then discard the remaining data.

src/btree/multi.ts
```diff
@@ -0,0 +1,78 @@
...
+		this.resolver = resolver;
+		this.offset = offset;
+	}
+
+	async root(): Promise<MemoryPointer | null> {
+		// we seek by 12 bytes since offset is 8 bytes, length is 4 bytes
+		const { data } = await this.resolver({
+			start: this.offset,
+			end: this.offset + 12,
+		});
```
kevmo314, line 18: Since the reads are more expensive in js, it's reasonable to have a `getPage()` call, something like:

```
private async getPage(): Promise<ArrayBuffer> {
  return await this.resolver({
    start: this.offset,
    end: this.offset + 4096 - 1,
  });
}
```

Cache this promise in the `LinkedMetaPage` class.

src/btree/multi.ts
```diff
@@ -0,0 +1,78 @@
...
+	constructor(resolver: RangeResolver, offset: number) {
+		this.resolver = resolver;
+		this.offset = offset;
+	}
+
+	async root(): Promise<MemoryPointer | null> {
+		// we seek by 12 bytes since offset is 8 bytes, length is 4 bytes
+		const { data } = await this.resolver({
+			start: this.offset,
+			end: this.offset + 12,
```
kevmo314, line 17: Be careful, I think in js the end offset is inclusive, so reading 12 bytes would be `this.offset + 12 - 1`.

src/btree/multi.ts
```diff
@@ -0,0 +1,78 @@
...
+			return null;
+		}
+
+		const view = new DataView(data);
+
+		const nextOffset = view.getBigUint64(0);
+
+		// since 2^64 - 1
+		const maxUint64 = BigInt(2) ** BigInt(64) - BigInt(1);
+		if (nextOffset === maxUint64) {
```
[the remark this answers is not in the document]
kevmo314: Yes this is correct, if the next pointer is set to maxUint64 it's the end.

src/btree/pagefile.ts
```diff
@@ -0,0 +1,44 @@
...
+		}
+
+		const start = pageIndex * this.pageSize;
+		const end = start + this.pageSize - 1;
+		const { data } = await this.resolver({ start, end });
+
+		return data;
+	}
+
+	async readFreePageIndex(): Promise<number[]> {
```
kevmo314, line 26: The js client doesn't need to worry about free pages, that's a writer detail.

kevmo314 (approved):
This looks pretty good, definitely in the right direction. Some optimization comments but otherwise it should work and the design looks fine.

src/btree/pagefile.ts
friendlymatthew: Sounds good

src/btree/multi.ts
friendlymatthew: > The first four bytes represent the length and then discard the remaining data.

I assume you mean retrieve the length from the first four bytes, then retrieve the metadata from:

```js
{ start: this.offset + 28, end: this.offset + 28 + metdatalength - 1 }
```

Discard the rest

kevmo314: Ah yes, however I'm saying that you can read it in one go. So I think in line with the other comment the idea is that when pulling a `LinkedMetaPage`, read all 4096 bytes at once then slice it up as you wish instead of calling the resolver multiple times for each desired range.

There's a slight benefit to doing it this way, namely the backend right now guarantees that each page is written atomically, so we want to read in pages too to guarantee we're not desynchronizing across different requests.

src/btree/multi.ts
```diff
@@ -0,0 +1,74 @@
...
+
+		const { data } = await this.resolver({
+			start: this.offset,
+			end: this.offset + 4096 - 1,
+		});
+
+		this.pageData = data;
+
+		return data;
+	}
```
friendlymatthew, line 57: we cache the pageData, all other methods in this class slice into pageData

src/btree/pagefile.ts
```diff
@@ -0,0 +1,25 @@
+import { RangeResolver } from "../resolver";
+
+const MAX_FREE_PAGE_INDICIES = 512;
+const PAGE_SIZE_BYTES = 4096;
+
+export class PageFile {
```
friendlymatthew, line 6: I don't think we need this class anymore, since we have `getPage()`

But I may be incorrect

PR #105, opened 2024-02-13
Refactor `btree.ts` and `node.ts`, starting coverage

friendlymatthew:
After merging the new btree structure, we also needed to make the following changes.


src/btree/node.ts
```diff
@@ -1,158 +1,139 @@
...
+		data: ArrayBuffer
+	): Promise<{ node: BPTreeNode; bytesRead: number }> {
+		const { data: bufferData } = await resolver({
+			start: Number(mp.offset),
+			end: Number(mp.offset) + 4096 - 1,
+		});
+		const node = new BPTreeNode([], [], [], data);
+		await node.unmarshalBinary(bufferData);
+
+		return { node, bytesRead: 4096 };
```
friendlymatthew, line 137: We pass in the `data: ArrayBuffer` from the `bptree`. Now that I think about it, since this is array is for our values, this should be a `Uint8Array`.

Also for `bytesRead`, I should probably return `m` from `unmarshalBinary`

src/tests/node.test.ts
```diff
@@ -36,189 +34,64 @@ describe("test compare bytes", () => {
...
-				totalLength: end - start,
-			};
-		});
+	it("reads from buffer for leaf node", async() => {
 
-		const { node, bytesRead } = await BPTreeNode.fromMemoryPointer(
-			nonLeafMemoryPointer,
-			mockRangeResolver
-		);
+		const buffer = await readBinaryFile('leaf_node_data.bin');
```
friendlymatthew, line 88: buffer output saved from the go test crate

kevmo314 (approved):
Nice! This looks really good

friendlymatthew:
> Nice! This looks really good

There is a lot more to come. Specifically mocking range resolvers and writing coverage :0

PR #106, opened 2024-02-14
fix: use offset as a disambiguator

kevmo314:
Fixes uniqueness issue


pkg/btree/bptree.go
```diff
@@ -40,19 +41,20 @@ func (t *BPTree) root() (*BPTreeNode, MemoryPointer, error) {
 	return root, mp, nil
 }
 
-func (t *BPTree) Find(key []byte) (MemoryPointer, bool, error) {
+func (t *BPTree) Find(key ReferencedValue) (ReferencedValue, MemoryPointer, error) {
```
friendlymatthew, line 44: Why are we returning the `ReferencedValue` in `Find()`?

Am I wrong to assume that the key `ReferencedValue` param will be returned?

friendlymatthew: > but the first key that matches could be []byte("hellp"), so returning the matched key provides a way to identify if the specific key searched exists or not.

Sounds good -- this means iff there exists a matched key, we'll return that key and it's `MemoryPointer
.

Will fix jsonl and csv tests

pkg/handlers/jsonl.go
```diff
@@ -185,10 +185,16 @@ func handleJSONLObject(f *appendable.IndexFile, r []byte, dec *json.Decoder, pat
...
 				}
 			case nil:
+				fmt.Printf("\n\ninserting data with offset: %v\n", data.Offset)
 				// nil values are a bit of a degenerate case, we are essentially using the btree
 				// as a set. we store the value as an empty byte slice.
-				if err := page.BPTree(r).Insert(btree.ReferencedValue{Value: []byte{}}, data); err != nil {
-					return fmt.Errorf("failed to insert into b+tree: %w", err)
+				if err := page.BPTree(r).Insert(btree.ReferencedValue{
+					Value:       []byte{},
+					DataPointer: data,
```
friendlymatthew, line 193: The `csv_test`, `jsonl_test`, and the `index-file` equality tests are passing with one specific exception.

The case that fails is when we have a null column, that is when:
```json
{ "header": null }
{ "header": null }
{ "header": null }
```. 

The above approach seems to fix that test case, but I was wondering if it was intentional to zero out the `DataPointer` attribute in `ReferencedValue`

Note when we compare bytes, the `key` offset and length are always zeroed.

PR #107, opened 2024-02-15
multi traverse test

friendlymatthew:
The new multi test case tests the `next()` function.

The `mockRangeResolver` makes the following assumptions:
1) fill the first 4096 bytes with zeros, since this is the garbage collector
2) for the subsequent `ith` page, it stores the `nextpointer` from the 12th byte position to 12 + 7th byte position. The next pointer is the offset of the start of the `i+1th` page.

### DIFF
--- a/src/btree/multi.ts
+++ b/src/btree/multi.ts
@@ -39,9 +39,8 @@ export class LinkedMetaPage {
    */
   async metadata(): Promise<ArrayBuffer> {
     const pageData = await this.getMetaPage();
-    const lengthData = pageData.slice(24, PAGE_SIZE_BYTES);
 
-    const lengthView = new DataView(lengthData);
+    const lengthView = new DataView(pageData, 24);
 
     // read the first four because that represents length
     const metadataLength = lengthView.getUint32(0);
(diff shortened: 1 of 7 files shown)


src/tests/multi.test.ts
```diff
@@ -1,51 +1,115 @@
...
+      }
+
+      return {
+        data: slice.buffer,
+        totalLength: view.byteLength,
+      };
+    };
+
+    const pageFile = new PageFile(mockRangeResolver);
+    const tree = ReadMultiBPTree(mockRangeResolver, pageFile);
```
friendlymatthew, line 86: Just want to point out that `ReadMultiBPTree` skips the `GC` page and initializes from the `fileMeta`. 

So in this case `tree` is actually the `fileMeta` page

src/btree/multi.ts
```diff
@@ -1,99 +1,101 @@
...
+
+  /**
+   * `metadata()` gets the page data. It does the following:
+   * 		(1) creates a slice from 24 to the end of the page
+   * 		(2) it reads the first four bytes of that slice which gives us the length to seek to
+   * 		(3) slices from [24, (24 + dataLength)] which contain metadata
+   */
+  async metadata(): Promise<ArrayBuffer> {
+    const pageData = await this.getMetaPage();
+    const lengthData = pageData.slice(24, PAGE_SIZE_BYTES);
```
kevmo314, line 42: The `DataView` constructor takes a byte offset and length so you don't need to slice the underlying data: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/DataView

src/btree/multi.ts
```diff
@@ -1,99 +1,101 @@
...
+  /**
+   * `next()` - returns a new LinkedMetaPage
+   */
+  async next(): Promise<LinkedMetaPage | null> {
+    const pageData = await this.getMetaPage();
+    const data = pageData.slice(12, 12 + 8);
+
+    const view = new DataView(data);
+    const nextOffset = view.getBigUint64(0);
+    const maxUint64 = BigInt(2) ** BigInt(64) - BigInt(1);
```
kevmo314, line 80:
```suggestion
    const maxUint64 = 2n ** 64n - 1n;
```

kevmo314 (approved):

PR #110, opened 2024-02-16
Update Field Type + traverse index headers

friendlymatthew:


src/index-file/meta.ts
```diff
@@ -23,54 +23,69 @@ export async function readFileMeta(buffer: ArrayBuffer): Promise<FileMeta> {
...
 export type IndexMeta = {
 	fieldName: string;
-	fieldType: bigint;
+	fieldType: number;
 };
 
-export async function unmarshalBinaryForIndexMeta(
-	resolver: RangeResolver,
-	buffer: ArrayBuffer
+export type IndexHeader = {
```
kevmo314, line 43: Is this all duplicating the code in the other file?

friendlymatthew: Yes but also `IndexHeader` is suppose to condense `IndexMeta[]` since for a given `IndexMeta`, there exists more than one element. `IndexHeader[]` ensures every `IndexHeader` has a unique `fieldName` and a list of it's `fieldType`.

src/btree/multi.ts
```diff
@@ -3,96 +3,106 @@ import { MemoryPointer } from "./node";
...
+		}
+
+		if (!this.metaPagePromise) {
+			this.metaPagePromise = this.resolver({
+				start: Number(this.offset),
+				end: Number(this.offset) + PAGE_SIZE_BYTES - 1,
+			})
+				.then(({ data }) => {
+					this.metaPageData = data;
+					this.metaPagePromise = null;
```
friendlymatthew, line 69: after a successful query, we set the pagePromise to null.

kevmo314 (approved):

PR #112, opened 2024-02-17
Bptree Test Coverage

friendlymatthew:
Refactored `binarySearchReferencedValue` to factor in offsets and length


src/tests/bptree.test.ts
```diff
@@ -1,49 +1,110 @@
...
+
+		const key1: ReferencedValue = new ReferencedValue(
+			{ offset: 0n, length: 0 },
+			new Uint8Array([1])
+		);
+
+		const [index1, found1] = binarySearchReferencedValues(values, key1);
+
+		expect(index1).toEqual(1);
+		expect(found1).toBeFalsy();
```
friendlymatthew, line 85: Just want to note that even though it successfully finds the index1, it still returns false.

src/tests/bptree.test.ts
```diff
@@ -1,49 +1,127 @@
...
 
-		expect(res).toEqual([{ offset: BigInt(0), length: 0 }, false]);
-		*/
+		const result = ReferencedValue.compareBytes(buffer1, buffer2);
+		console.log(result); 
+
+    expect(result).toEqual(0)
   });
+
+	it("should compare reference values", () => {
```
friendlymatthew, line 80: I've asserted that this function behaves accordingly by writing out the exact test coverage for the go implementation.

`Found` will only return `true` iff the `value` exists and the `MemoryPointer` is that exact one.

src/tests/bptree.test.ts
```diff
@@ -1,49 +1,102 @@
...
-		expect(res).toEqual([{ offset: BigInt(0), length: 0 }, false]);
-		*/
-  });
+			if (rv && rv.value) {
+				const rvValue = new DataView(rv.value).getBigUint64(0);
+				console.log("r: ", rvValue);
+			} else {
+				console.log("value undefined: ", idx);
+				break;
+			}
```
friendlymatthew, line 99: for this test case, it seeks and finds upto `idx: 85`. This is because the given root node is from values [0 to 84]. My current intuition is that the first node's page has been reached and it's having a hard time jumping to the next page.

kevmo314: Yep that's correct, each page stores ~100 values and if you're storing ints like in the test the first page will split and end at ~85.

kevmo314: To jump to the next page you'll have to implement the `.Next()` iterator function to do the traversal.

friendlymatthew: > To jump to the next page you'll have to implement the `.Next()` iterator function to do the traversal.

Do you mean something like: https://github.com/kevmo314/appendable/pull/112/files#diff-0efa68a1779c18545bc009ae1e22699bd9df18732220296ea2e2a6073b3b5d11R45-R101

If this is the right approach, I'll start debugging why it's not behaving accordingly

kevmo314: Ah yeah that's right, but in this test you're not calling that iterator in this test. Here's how to call it: https://github.com/kevmo314/appendable/blob/main/pkg/btree/bptree_test.go#L341-L362

friendlymatthew: I thought the `find()` would internally call the iterator: https://github.com/kevmo314/appendable/pull/112/files#diff-2ebc6cac248e11e45734206a7f4f04550ca2b42269743a8a5143b11320eb9072R98-R115

But I may be incorrect.

pkg/btree/bptree_test.go
```diff
@@ -221,10 +232,15 @@ func TestBPTree_SequentialInsertionTest(t *testing.T) {
 	for i := 0; i < 256; i++ {
 		buf := make([]byte, 8)
 		binary.BigEndian.PutUint64(buf, uint64(i))
-		if err := tree.Insert(ReferencedValue{Value: buf}, MemoryPointer{Offset: uint64(i)}); err != nil {
+		if err := tree.Insert(ReferencedValue{Value: buf}, MemoryPointer{Offset: uint64(i), Length: uint32(len(buf))}); err != nil {
 			t.Fatal(err)
 		}
 	}
+
+	if err := b.WriteToDisk("bptree_sequential.bin"); err != nil {
```
kevmo314, line 240: It would be useful to make this a separate entrypoint. You can make a `test/sequential/main.go` or something to generate the binary file. That way the entire test suite doesn't have to be run to generate the file and the tests stay side-effect-free.

pkg/btree/bptree_test.go
```diff
@@ -247,6 +263,57 @@ func (s *StubDataParser) Parse(value []byte) []byte {
 	return []byte{1, 2, 3, 4, 5, 6, 7, 8}
 }
 
+func TestBinarySearchReferencedValues(t *testing.T) {
```
kevmo314, line 266: This test isn't really testing any of our code since it calls `slices.BinarySearchFunc`. It might be useful to leave as an example but I don't think we should run it with unit tests. Name it `func Example*()` if you want to leave it in: https://go.dev/blog/examples

pkg/btree/bptree_test.go
```diff
@@ -338,6 +407,10 @@ func TestBPTree_Iteration(t *testing.T) {
 		}
 	}
 
+	b.WriteToDisk("btree_iterator.bin")
```
kevmo314, line 410: Same here, I think this would be better suited to a separate entrypoint.

pkg/buftest/buffer.go
```diff
@@ -86,6 +89,14 @@ func (b *SeekableBuffer) ReadAt(p []byte, off int64) (int, error) {
 	return n, nil
 }
 
+func (b *SeekableBuffer) WriteToDisk(filename string) error {
+	if err := os.WriteFile(filename, b.buf, 0644); err != nil {
+		return err
+	}
+
+	return nil
```
kevmo314, line 97:
```suggestion
	return os.WriteFile(filename, b.buf, 0644)
```

src/btree/traversal.ts
```diff
@@ -0,0 +1,100 @@
...
+		const rolloverRight =
+			this.records[i].index >= this.records[i].node.numPointers();
+
+		if (rolloverLeft || rolloverRight) {
+			if (!this.increment(i + 1, delta)) {
+				return false;
+			}
+
+			try {
+				const node = await this.tree.readNode(
```
kevmo314, line 63: It would be fine to let this error propagate, it doesn't have to be caught.

src/tests/bptree_1.bin
kevmo314, line 1: Don't check in these binary files, they should be generated in a `beforeAll()` statement or something. Checking in binary files makes cloning slower and it's not super easy to recover from.

kevmo314 (commented):
Remove the `fmt.Printf()` calls, it would be better to keep the log clean. stdout is not the most performant, so adding stdout log statements slows the entire processing down.

This looks pretty good overall though, just some cleaning up.

src/btree/traversal.ts
friendlymatthew: That was my initial intuition, but in order for the iterator to properly halt, we need to cast the error as a `false` value. 

The error that propagates is an undefined indexing error, which means when it reaches the first `out of bounds` error, we can cast it as `false` and essentially halt the iterator.

kevmo314: Which test is this during? `readNode()` should only error if the node itself can't be read due to network issues?

friendlymatthew: > Which test is this during? `readNode()` should only error if the node itself can't be read due to network issues?

The iterator test cases in `bptree.test.ts` errs when we remove the `try {} catch {}` block. 

It errs: 
```sh
TypeError: Cannot read properties of undefined (reading 'node')

      58 |
      59 |                      const node = await this.tree.readNode(
    > 60 |                              this.records[i + 1].node.pointer(this.records[i + 1].index)
         |                                                  ^
      61 |                      );
      62 |
      63 |                      // propagate the rollover
```

This is because `this.records[i + 1]` is undefined. I removed the block and caught it earlier on in `increment`

src/btree/traversal.ts
```diff
@@ -0,0 +1,99 @@
...
+			// propagate the rollover
+			this.records[i].node = node;
+
+			if (rolloverLeft) {
+				this.records[i].index = this.records[i].node.numPointers() - 1;
+			} else {
+				this.records[i].index = 0;
+			}
+
+			// otherwise, update the current node
```
kevmo314, line 75:
```suggestion
```

kevmo314 (approved):

PR #126, opened 2024-02-22
Bug: iterator not iterating sequentially

friendlymatthew:
Upon getting the `first()` and `last()` values in a `bptree`, we want to iterate and make sure the next value is non-decreasing or non-increasing.


pkg/mocks/mock.index
kevmo314, line 1: Don't check this file in

pkg/btree/bptree_test.go
```diff
@@ -470,3 +471,98 @@ func TestBPTree_Iteration_SinglePage(t *testing.T) {
...
+		binary.LittleEndian.PutUint64(buf, math.Float64bits(i))
+
+		if err := tree.Insert(ReferencedValue{Value: buf}, MemoryPointer{Offset: uint64(i * 100), Length: uint32(len(buf))}); err != nil {
+			t.Fatal(err)
+		}
+
+		if int(i*100)%10 == 0 && currentIncrementIndex < len(increments)-1 {
+			currentIncrementIndex++
+		}
+	}
```
friendlymatthew, line 497: to sort of mimic the dataset we're dealing with, I inserted a bunch of different floats, varying with decimal size. 

However, as we iterate, the value we insert should always be increasing

PR #130, opened 2024-02-23
feat: query validation

friendlymatthew:
Validates the query before firing


src/db/query-validation.ts
```diff
@@ -170,12 +165,12 @@ function validateSelect<T extends Schema>(
  * @throws {Error} Throws an error if query is invalid.
  */
 export async function validateQuery<T extends Schema>(
```
kevmo314, line 167: This function doesn't have to be async

kevmo314 (approved):

PR #137, opened 2024-02-26
feat: accept multiple ranges

### LINKED ISSUE #80: Make `RangeResolver` accept a list of ranges
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range

> sometimes the b+ tree doesn't store the actual value, it stores a pointer. The js version will need to handle that but more importantly, when resolving a node we should make one http request with all the relevant ranges instead of resolving once for each data reference pointer in the b+tree node

friendlymatthew:
Fixes #80

### DIFF
--- a/src/btree/multi.ts
+++ b/src/btree/multi.ts
@@ -60,11 +60,14 @@ export class LinkedMetaPage {
     }
 
     if (!this.metaPagePromise) {
-      this.metaPagePromise = this.resolver({
-        start: Number(this.offset),
-        end: Number(this.offset) + PAGE_SIZE_BYTES - 1,
-      })
-        .then(({ data }) => {
+      this.metaPagePromise = this.resolver([
+        {
+          start: Number(this.offset),
+          end: Number(this.offset) + PAGE_SIZE_BYTES - 1,
+        },
+      ])
+        .then((r) => {
+          const { data } = r[0];
           this.metaPageData = data;
           this.metaPagePromise = null;
           return data;
@@ -101,7 +104,7 @@ export class LinkedMetaPage {
 
 export function ReadMultiBPTree(
   resolver: RangeResolver,
-  pageFile: PageFile
+  pageFile: PageFile,
 ): LinkedMetaPage {
   const offset = pageFile.page(0);
   return new LinkedMetaPage(resolver, offset);
--- a/src/btree/multi.ts
+++ b/src/btree/multi.ts
@@ -5,20 +5,22 @@ import { PageFile } from "./pagefile";
 const PAGE_SIZE_BYTES = 4096;
 export const maxUint64 = 2n ** 64n - 1n;
 
+const globalMetaPageCache = new Map<
+  bigint,
+  Promise<{ data: ArrayBuffer; totalLength: number }[]>
+>();
+
 export class LinkedMetaPage {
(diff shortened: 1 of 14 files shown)


src/btree/node.ts
```diff
@@ -140,6 +143,19 @@ export class BPTreeNode {
...
+      res.forEach((res, index) => {
+        const dpIndex = dpIndexes[index];
+        const { data } = res;
+        console.log("data: ", data);
+
+        const parsedData = this.parseValue(data);
+        this.keys[dpIndex].setValue(parsedData);
+      });
+    }
+
```
friendlymatthew, line 158: this was the only time I really used the ranges to resolve them at once

src/data-file.ts
```diff
@@ -1,46 +1,108 @@
...
+        throw new Error(`Expected 206 Partial Content, got ${response.status}`);
+      }
+
+      const contentType = response.headers.get("Content-Type");
+      if (!contentType) {
+        throw new Error("Missing Content-Type in response");
+      }
+      if (contentType.includes("multipart/byteranges")) {
+        const boundary = contentType.split("boundary=")[1];
+        const abuf = await response.arrayBuffer();
```
kevmo314, line 55:
```suggestion
        const text = await response.text();
```

Does this work?

src/data-file.ts
```diff
@@ -1,46 +1,108 @@
...
+        const text = new TextDecoder().decode(abuf);
+
+        const chunks = parseMultipartBody(text, boundary);
```
kevmo314, line 58:
```suggestion
        const chunks = parseMultipartBody(text, boundary.slice(0, boundary.length - 1));
```

I think this would avoid needing to pop?

src/data-file.ts
```diff
@@ -1,46 +1,108 @@
...
+
+        // the last element is null since the final boundary marker is followed by another delim.
+        if (chunks[chunks.length - 1].body === undefined) {
+          chunks.pop();
+        }
+        console.log(chunks);
+        return chunks.map((c) => {
+          const enc = new TextEncoder();
```
kevmo314, line 66: Use the same encoder across all chunks.

src/index-file/index-file.ts
```diff
@@ -9,41 +9,75 @@ import {
...
-        return {
-          data: await response.arrayBuffer(),
-          totalLength: totalLength,
-        };
+        return [
+          {
+            data: abuf,
+            totalLength: totalLength,
+          },
+        ];
```
kevmo314, line 67: Can this be put into a shared function with `DataFile`? Maybe create a `range-requests.ts` and export the function to parse the response.

kevmo314 (commented):
Looks good overall! Just a few simplifications.

src/btree/multi.ts
```diff
@@ -5,20 +5,22 @@ import { PageFile } from "./pagefile";
 const PAGE_SIZE_BYTES = 4096;
 export const maxUint64 = 2n ** 64n - 1n;
 
+const globalMetaPageCache = new Map<
```
friendlymatthew, line 8: resolved all comments above -- added global meta page cache.

It caches properly now

kevmo314 (approved):

PR #141, opened 2024-02-27
feat: Skiplist data structure to expedite meta page searching

### LINKED ISSUE #138: Implement a skip list for meta page pointers
Right now, each meta page only points at the next page which means fetching all the pages requires O(n) http requests.

We can improve this by implementing a skip list-type datastructure in the meta page sequence, taking advantage of the fact that most metadata isn't really used anyways. https://en.wikipedia.org/wiki/Skip_list

Note, instead of doing this to improve seeking performance, we really just want to store the next n pointers.

This isn't so easy to do trivially with the current API, so it probably makes sense to add an optimization pass API that will go through and update the meta page pointers and call that periodically.

friendlymatthew:
closes #138 

## Skiplist details
1. `coinFlip()` is a probabilistic balancing feature 
2. We treat the `IndexMeta` as the `SkipListNode's` composite key

- [x] wrote tests
- [x] counted searching time for demo. Variance where the quickest was 2 iterations and the average was ~16/2

### DIFF
--- a/src/index-file/skiplist.ts
+++ b/src/index-file/skiplist.ts
@@ -102,4 +102,8 @@ export class SkipList {
       `metapage not found for ${query.fieldName} and ${query.fieldType}`,
     );
   }
+
+  getSentinel(): SkipListNode {
+    return this.head;
+  }
 }
--- a/src/index-file/skiplist.ts
+++ b/src/index-file/skiplist.ts
@@ -1,0 +1,105 @@
+import { LinkedMetaPage } from "../btree/multi";
+import { FieldType } from "../db/database";
+import { IndexMeta, readIndexMeta } from "./meta";
+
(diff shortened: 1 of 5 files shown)


src/index-file/skiplist.ts
```diff
@@ -0,0 +1,109 @@
...
+    this.head = {
+      fieldName: "",
+      fieldType: FieldType.String, // since this value is 0
+      mp: null,
+      next: new Array(this.maxLevel).fill(null),
+    };
+  }
+
+  private coinFlip(): number {
+    const p = 0.5; // the probability factor
```
friendlymatthew, line 27: this was from the wikipedia implementation details

friendlymatthew:
Closing as not planned

PR #146, opened 2024-02-28
Store the next N offsets

### LINKED ISSUE #138: Implement a skip list for meta page pointers

friendlymatthew:
Closes #138


src/tests/mock_binaries/rangeresponse.bin
friendlymatthew, line 1: this is the range response that I captured. 

I did the following to export it:
```ts

    const blob = await response.blob();
    const url = window.URL.createObjectURL(blob);
```

src/multipart.ts
```diff
@@ -100,6 +103,7 @@ export default async function* parseMultipartBody(
         }
       }
       if (buf[ptr] !== boundary.charCodeAt(i)) {
```
friendlymatthew, line 105: this fails right now because buffer is initialized, but the contents are undefined

pkg/btree/multi.go
```diff
@@ -97,6 +99,60 @@ func (m *LinkedMetaPage) MarshalMetadata(bm encoding.BinaryMarshaler) error {
 	return m.SetMetadata(buf)
 }
 
+func (m *LinkedMetaPage) NextNOffsets() ([]uint64, error) {
```
kevmo314, line 102: Hmm I see what you're getting at here but I wouldn't expect this to actually traverse the linked list.

Let's pass the offsets into `SetNextNOffsets(offsets []uint64)` instead, and in the caller manage the list of offsets. In other words, make the `NextNOffsets` memory block "dumb" in that it only stores a list of up to 16 items. That way this API stays consistent with the other setters and getters.

pkg/btree/multi.go
```diff
@@ -97,6 +99,60 @@ func (m *LinkedMetaPage) MarshalMetadata(bm encoding.BinaryMarshaler) error {
...
+	}
+
+	for _, offset := range offsets {
+		if err := binary.Write(m.rws, binary.LittleEndian, offset); err != nil {
+			return err
+		}
+	}
+
+	for i := len(offsets); i < N; i++ {
+		if err := binary.Write(m.rws, binary.LittleEndian, ^uint64(0)); err != nil {
```
kevmo314, line 149: Nice this is pretty clever. One improvement: since you stop iterating after the first ending offset, only one offset has to be set to 0xFFFFFFFF. The rest can be left as whatever's in the buffer.

src/range-request.ts
```diff
@@ -1,23 +1,98 @@
...
+          if (expectedLength && totalLength !== expectedLength) {
+            throw new LengthIntegrityError();
+          }
+          return {
+            data: await res.arrayBuffer(),
+            totalLength: totalLength,
+          };
+        },
+      );
+      return Promise.all(individualRangePromises)
```
kevmo314, line 41:
```suggestion
      return await Promise.all(individualRangePromises);
```

pkg/appendable/index_file.go
```diff
@@ -192,6 +192,37 @@ func (i *IndexFile) FindOrCreateIndex(name string, fieldType FieldType) (*btree.
...
+
+		offsets, err := mp.NextNOffsets()
+		if err != nil {
+			return err
+		}
+
+		err = mp.SetNextNOffsets(offsets)
+		if err != nil {
+			return err
+		}
```
friendlymatthew, line 219: comments have been resolved 👍

pkg/btree/multi.go
```diff
@@ -97,6 +99,52 @@ func (m *LinkedMetaPage) MarshalMetadata(bm encoding.BinaryMarshaler) error {
 	return m.SetMetadata(buf)
 }
 
+func (m *LinkedMetaPage) NextNOffsets() ([]uint64, error) {
+	offsets := make([]uint64, N)
+	currentOffset := m.offset
+
+	for i := 0; i < N; i++ {
+		if _, err := m.rws.Seek(int64(currentOffset)+12, io.SeekStart); err != nil {
```
kevmo314, line 107: Ahh I mean I don't think this code should iterate through offsets.

As in, this entire function should be

```
Seek(currentOffset + 12)
for i := 0; i < N; i++ {
  binary.Read(rws, binary.LittleEndian, &offset[i]);
}
```

It should only read from the current meta page.

kevmo314 (approved):

PR #151, opened 2024-03-01
Read inline values

### LINKED ISSUE #135: Improve Value + DataPointer storage scheme
Right now, we store either the value or DataPointer but because we use the DataPointer as a disambiguator, it's impossible to not store the DataPointer, causing values to not be possible to inline. We should instead update the scheme to store both the value and DataPointer when a lookup is not desirable. We probably want a heuristic to figure out if the value should be stored inline.

friendlymatthew:
Closes #135 

## Some metrics:
Screen Shot 2024-03-02 at 11 52 15 PM

For the demo dataset, we make zero http requests to parse the value from a key's memory pointer. Everything gets read inline from disk.

Screen Shot 2024-03-02 at 11 26 36 PM

For one query, we make 10 requests **total**.

- [x] wrote test coverage


kevmo314 (approved):
Good stuff 🚀🚀🚀

PR #155, opened 2024-03-03
chore: delete protocol directory

### LINKED ISSUE #153: Remove protocol package
I think this isn't used anymore? https://github.com/kevmo314/appendable/blob/main/pkg/protocol/protocol.go

kevmo314:
Closes #153


friendlymatthew (approved):
Just checked the ts library and everything seems to be geared towards the btree implementation

PR #162, opened 2024-03-04
feat: add useMultipart flag

### LINKED ISSUE #158: Add flag to enable/disable multipart/byteranges
It seems like relying on a 200 response is too late if the server doesn't send multiple ranges: 
image

Let's add an API to the Appendable constructor to disable it, something like `Appendable.init(...,..., { useMultipartByteRanges: false })` and if we get a 200, warn the user if this flag is not set.

friendlymatthew:
Closes #158 

In our `init()`, this PR introduces a `Config` which contains the following:

```ts
export type Config = {
     useMultipartByteRanges?: boolean
}
```

We drill this through and in `range-request.ts`, we add logic where if that flag is either undefined or false, we'll resolve each range individually, else we'll continue with multi-fetching. If we reach a `200` response when multi-fetching, we'll warn the user that this flag hasn't been set.

Tested via local

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,8 +10,12 @@ export type Config = {
 export async function init(
   dataUrl: string | RangeResolver,
   indexUrl: string | RangeResolver,
-  config: Config,
+  config?: Config,
 ) {
+  if (!config) {
+    config = { useMultipartByteRanges: true };
+  }
+
   return Database.forDataFileAndIndexFile(
     typeof dataUrl === "string"
       ? DataFile.forUrl(dataUrl, config)
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,16 +3,21 @@ import { Database, FieldType, fieldTypeToString } from "./db/database";
 import { IndexFile } from "./index-file/index-file";
 import { RangeResolver } from "./resolver";
 
+export type Config = {
+  useMultipartByteRanges?: boolean;
+};
+
 export async function init(
   dataUrl: string | RangeResolver,
   indexUrl: string | RangeResolver,
+  config: Config,
 ) {
   return Database.forDataFileAndIndexFile(
     typeof dataUrl === "string"
-      ? DataFile.forUrl(dataUrl)
+      ? DataFile.forUrl(dataUrl, config)
       : DataFile.forResolver(dataUrl),
     typeof indexUrl === "string"
-      ? await IndexFile.forUrl(indexUrl)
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -3,16 +3,21 @@ import { Database, FieldType, fieldTypeToString } from "./db/database";
...
 import { RangeResolver } from "./resolver";
 
+export type Config = {
+  useMultipartByteRanges?: boolean;
+};
+
 export async function init(
   dataUrl: string | RangeResolver,
   indexUrl: string | RangeResolver,
+  config: Config,
```
kevmo314, line 13:
```suggestion
  config?: Config = {},
```

src/range-request.ts
```diff
@@ -1,10 +1,48 @@
+import { Config } from ".";
 import parseMultipartBody from "./multipart";
 import { LengthIntegrityError } from "./resolver";
 
+async function resolveIndividualPromises(
+  url: string,
+  ranges: { start: number; end: number; expectedLength?: number }[],
+) {
+  console.log("resolving ranges individually");
```
kevmo314, line 9:
```suggestion
```

src/range-request.ts
```diff
@@ -1,10 +1,48 @@
...
+
 export async function requestRanges(
   url: string,
   ranges: { start: number; end: number; expectedLength?: number }[],
+  config: Config,
 ): Promise<{ data: ArrayBuffer; totalLength: number }[]> {
+  const { useMultipartByteRanges } = config;
+  if (
+    useMultipartByteRanges === false ||
+    useMultipartByteRanges === undefined
```
kevmo314, line 41:
```suggestion
    useMultipartByteRanges === false
```

The default behavior should be true

src/range-request.ts
```diff
@@ -18,28 +56,10 @@ export async function requestRanges(
...
-          if (expectedLength && totalLength !== expectedLength) {
-            throw new LengthIntegrityError();
-          }
-          return {
-            data: await res.arrayBuffer(),
-            totalLength: totalLength,
-          };
-        },
+      console.warn(
+        `useMultipartByteRanges has not been set to false. The server can not handle multipart byte ranges.`,
```
kevmo314, line 60:
```suggestion
        `useMultipartByteRanges is enabled but the server indicated did not respond with a subset of bytes. Set useMultipartByteRanges: false in your Appendable config object.`,
```

kevmo314 (approved):

PR #165, opened 2024-03-06
Set fixed width in `IndexMeta`

### LINKED ISSUE #152: Further improve value storage

* 1 byte for boolean
* 0 bytes for null

friendlymatthew:
Closes #152 

> [!NOTE]
> Still a draft PR. We'll need to update `node.go` and `bptree.go` to pass the `FixedWidth` information and remove the need of storing `l`

## What this PR contains

1. Allocates 2 bytes in the `IndexMeta` to specify the `FixedWidth` for that given multi page
2. Updated the `ts` reader so it reads the `FixedWidth` upon scanning the metadata, drills down to the `UnmarshalNode` logic, replacing the need to store the length (`l`) in disk

- [x] tested and works via demo

### DIFF
--- a/pkg/appendable/index_file.go
+++ b/pkg/appendable/index_file.go
@@ -185,10 +185,16 @@ func (i *IndexFile) FindOrCreateIndex(name string, fieldType FieldType) (*btree.
 	metadata := &IndexMeta{}
 	metadata.FieldName = name
 	metadata.FieldType = fieldType
+	fieldWidth := DetermineFieldWidth(fieldType)
+	metadata.FieldWidth = fieldWidth
+
 	buf, err := metadata.MarshalBinary()
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal metadata: %w", err)
 	}
+
+	next.FixedWidth = fieldWidth
+
 	return next, next.SetMetadata(buf)
 }
 
--- a/pkg/appendable/appendable.go
+++ b/pkg/appendable/appendable.go
@@ -122,27 +122,43 @@ func (m *FileMeta) UnmarshalBinary(buf []byte) error {
 }
 
 type IndexMeta struct {
-	FieldName string
-	FieldType FieldType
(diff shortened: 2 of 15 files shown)


kevmo314:
Nice work so far! Just a note, it's fine for the js API but for the go API we shouldn't add `fieldWidth` to `UnmarshalBinary` to avoid it from violating the `BinaryUnmarshaller` interface.

This admittedly makes it a little tricky but the way we could get around this is manually setting the field width on the struct before unmarshalling.

```
node := &BPTreeNode{ FieldWidth: ... }
node.UnmarshalBinary(...)
```

friendlymatthew:
> Nice work so far! Just a note, it's fine for the js API but for the go API we shouldn't add `fieldWidth` to `UnmarshalBinary` to avoid it from violating the `BinaryUnmarshaller` interface.
> 
> ```
> node := &BPTreeNode{ FieldWidth: ... }
> node.UnmarshalBinary(...)
> ```

Sounds good. I don't think my existing work violates the `UnmarshalBinary` logic for `node.go`. But I'll find an entry point and drill it down to `BpTreeNode`

pkg/appendable/index_file.go
```diff
@@ -185,10 +185,16 @@ func (i *IndexFile) FindOrCreateIndex(name string, fieldType FieldType) (*btree.
...
 	metadata.FieldType = fieldType
+	fieldWidth := DetermineFieldWidth(fieldType)
+	metadata.FieldWidth = fieldWidth
+
 	buf, err := metadata.MarshalBinary()
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal metadata: %w", err)
 	}
+
+	next.FixedWidth = fieldWidth
```
friendlymatthew, line 196: this should correctly store the `FixedWidth` in `LinkedMetaPage` right?

friendlymatthew:
I'm going to close this -- and start new

PR #174, opened 2024-03-08
feat: spec width in bptree

### LINKED ISSUE #152: Further improve value storage

friendlymatthew:
Closes #152


friendlymatthew:
Too many conflicts to resolve -- it's just much easier to clean slate this PR

PR #175, opened 2024-03-11
feat: spec width in bptree

### LINKED ISSUE #152: Further improve value storage

friendlymatthew:
Closes #152 

`DetermineWidth(ft FieldType)` determines the width, where `0` represents variable width and `1 + x` represents the fixed width `x`.

We drill this through the `bptree` lifecycle, updating how we write and read from disk.

We store the width in `IndexMeta` since we currently following a singleton relationship between a `MetaPage` and `Bptree`. 

Read accordingly to typescript.

- [x] fixed tests
- [x] behaves accordingly on demo


src/tests/mock_binaries/green_tripdata_2023-01.index
kevmo314, line 1: Does this have to be checked in?

kevmo314 (approved):

src/tests/mock_binaries/green_tripdata_2023-01.index
friendlymatthew: We use this file in `index-file.test` to read the file meta, and make sure we can traverse and collect the pages correctly

kevmo314: Can we generate it instead? It seems like an integration test which is good to have but having fixtures will be a maintenance burden later if we update the schema.

friendlymatthew: Yeah that makes sense -- I'll update accordingly

PR #176, opened 2024-03-11
feat: floating int version

### LINKED ISSUE #154: Use pessimistic length for DataPointer
There is an implementation of a pessimistic length here: https://github.com/kevmo314/appendable/blob/main/pkg/encoding/fint.go

We define a type fint16 as a _pessimistic_ uint16. For DataPointers, we can use fint16 for the length instead of a uint16. That is, let's update the data pointer to be

```
type DataPointer struct {
  Offset uint64
  Length FloatingInt16
}
```

to save 2 bytes per pointer.
 
The key idea behind a pessimistic length is that when passing a data pointer, it's sufficient for the length to be at least greater than the correct length. That is, suppose we have the data:

```
{"a":"test123","b":"whatever moo"}
```

The DataPointer for `a` is `{Offset: 5, Length: 9}`. When we resolve this in the data file, it resolves as `"test123"`. However, note that when we resolve it, it's actually ok if we were to try to resolve `"test123","b":"wha`, the resolver can be written such that it's smart enough to realize that `"test123"` is the actual string and everything else after it is chopped off. (try this with `JSON.parse('"test123","b":"wha'`)` in your browser).

So, in this case, it's actually ok if we had a data pointer that is `{Offset: 5, Length: 18}` instead of `Length: 9`. Floating ints try to take advantage of this: a floating int decodes accurately at lower ints (in other words, `uint32(fint16(uint32(1))) == uint32(1)` and `uint32(fint16(uint32(4))) == uint32(4)` and so on) but at higher values, `fint16` _overestimates_ the correct value (in other words, `uint32(fint16(uint32(1 << 16))) > uint32(1 << 16)` observing that at smaller values we want to be accurate but at larger values it's fine to overestimate as they're less likely to occur anyways. See the tests for more details: https://github.com/kevmo314/appendable/blob/main/pkg/encoding/fint_test.go

Therefore, we should be able to swap `Length` to be a `FloatingInt16` and save 2 bytes per data pointer.

friendlymatthew:
closes #154 

This is just the `FloatingInt16` version. The index file is `81.6MB`.


friendlymatthew:
Closing since #179 won

Issue #177, opened 2024-03-11
bug: iterator not halting as expected


friendlymatthew:
To reproduce:

When querying the following:

```json
 {
      "where": [
        {
          "operation": "==",
          "key": "trip_distance",
          "value": 10,
        },
      ],
      "orderBy": [
        {
          "key": "trip_distance",
          "direction": "ASC",
        },
      ]
}
```

we expect 10 results to appear. However, in `appendResult()`, we have an edge case that checks if the next query is done. If so, we break out of that loop. 

That is not working as expected, and we proceed with:
<img width="1315" alt="Screen Shot 2024-03-11 at 5 27 11 PM" src="https://github.com/kevmo314/appendable/assets/38759997/f016eac4-689c-483f-bfa4-d038e9481dd2">

friendlymatthew:
Closed via #184

PR #184, opened 2024-03-14
Feat: Fix iterator

friendlymatthew:
For https://github.com/kevmo314/appendable/issues/177

This test seeks to insert the same values, and iterate through + see how many times we cycle through. The demo behaves like we have one extra node in our searching


friendlymatthew:
So the good news is that the iterator correctly fetches the count 10 times. (See the 1023 btree test). 

This indicates the bug in #177 is purely a javascript issue.

PR #193, opened 2024-03-20
refactor: pager

### LINKED ISSUE #159: Pack multiple meta pages into a single page
Right now, each meta page takes up 4096 bytes. This is kind of a waste since meta pages, unlike internal nodes, will rarely take up that much space. Let's pack multiple meta pages into a single disk page, let's set the size of a meta page to 256 bytes. This will probably involve decoupling the "linked list" part of the meta page from the actual "page" part of the meta page so will need some memory and API rejiggering.

friendlymatthew:
closes #159


pkg/pagefile/pagefile.go
```diff
@@ -118,7 +128,7 @@ func (pf *PageFile) FreePageIndex() (int64, error) {
 	return offset, nil
 }
 
-func (pf *PageFile) NewPage(buf []byte) (int64, error) {
+func (pf *PageFile) NewPage(buf []byte, freeSlotIndexes *[][]bool) (int64, error) {
```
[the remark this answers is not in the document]
kevmo314: Hmm, what's the motivation for keeping track of the number of slots in `PageFile`? This seems potentially ok but I'm not sure about the necessity.

pkg/metapage/multi.go
```diff
@@ -0,0 +1,213 @@
...
+		return 0, err
+	}
+
+	pageIndex := newPageOffset / int64(m.rws.PageSize())
+	metaSlotsPerPage := m.rws.PageSize() / m.rws.SlotSize()
+	metaSlotsRow := make([]bool, metaSlotsPerPage)
+	metaSlotsRow[0] = true
+	m.freeSlotIndexes = append(m.freeSlotIndexes, metaSlotsRow)
+
+	return pageIndex * int64(m.rws.PageSize()), nil
```
friendlymatthew, line 153: I agree -- this logic should be enough for opening a new row of slots for the next subsequent page. 

I was trying to figure out why the BPtree errors out

pkg/btree/bptree.go
```diff
@@ -3,31 +3,31 @@ package btree
 import (
 	"bytes"
 	"fmt"
+	"github.com/kevmo314/appendable/pkg/common"
```
kevmo314, line 6: `common` isn't a great name for a package, by definition everything is common if it's in a package. :)

Maybe `ptr` or something?

kevmo314 (approved):

PR #195, opened 2024-03-20
Replace `%` for `&` + Benchmarks

friendlymatthew:
In `PageFile`, there are a lot of modulo computations. However, if the divisor is a base of 2, you can do the following:

`m % n === m & (n-1)`. [According to this blogpost](https://mziccard.me/2015/05/08/modulo-and-division-vs-bitwise-operations/), modulo takes multiple CPU cycles while bitwise only cycles once. 

Here are the benchmarks:

- modulo
```
goos: darwin
goarch: arm64
pkg: github.com/kevmo314/appendable/pkg/pagefile
BenchmarkPageFile
BenchmarkPageFile/Benchmark_new_page_file
BenchmarkPageFile/Benchmark_new_page_file-10         	 1546320	       774.8 ns/op
BenchmarkPageFile/Benchmark_writeFreePageIndices
BenchmarkPageFile/Benchmark_writeFreePageIndices-10  	 3029023	       396.1 ns/op
BenchmarkPageFile/Benchmark_free_page_index
BenchmarkPageFile/Benchmark_free_page_index-10       	586616586	         2.045 ns/op
BenchmarkPageFile/Benchmark_new_page
BenchmarkPageFile/Benchmark_new_page-10              	  495770	      2442 ns/op
```
<br>

- this is with bitwise

```
goos: darwin
goarch: arm64
pkg: github.com/kevmo314/appendable/pkg/pagefile
BenchmarkPageFile
BenchmarkPageFile/Benchmark_new_page_file
BenchmarkPageFile/Benchmark_new_page_file-10         	 1535476	       769.8 ns/op
BenchmarkPageFile/Benchmark_writeFreePageIndices
BenchmarkPageFile/Benchmark_writeFreePageIndices-10  	 2955009	       404.6 ns/op
BenchmarkPageFile/Benchmark_free_page_index
BenchmarkPageFile/Benchmark_free_page_index-10       	585331350	         2.050 ns/op
BenchmarkPageFile/Benchmark_new_page
BenchmarkPageFile/Benchmark_new_page-10              	  471688	      2182 ns/op
```

### DIFF
--- a/pkg/pagefile/pagefile.go
+++ b/pkg/pagefile/pagefile.go
@@ -57,7 +57,7 @@ func NewPageFile(rws io.ReadWriteSeeker) (*PageFile, error) {
 			offset := int64(binary.LittleEndian.Uint64(buf[i*8 : (i+1)*8]))
 			if offset != 0 {
 				pf.freePageIndexes[pf.freePageHead] = offset
-				pf.freePageHead = (pf.freePageHead+1)&len(pf.freePageIndexes) - 1
+				pf.freePageHead = (pf.freePageHead + 1) & (len(pf.freePageIndexes) - 1)
 				pf.freePageCount++
 			} else {
 				break
--- a/pkg/pagefile/pagefile.go
+++ b/pkg/pagefile/pagefile.go
@@ -57,7 +57,7 @@ func NewPageFile(rws io.ReadWriteSeeker) (*PageFile, error) {
 			offset := int64(binary.LittleEndian.Uint64(buf[i*8 : (i+1)*8]))
 			if offset != 0 {
 				pf.freePageIndexes[pf.freePageHead] = offset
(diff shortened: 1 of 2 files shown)


friendlymatthew (commented):
Seems like the go compiler already performs this optimization for divisors that are powers of 2?

https://godbolt.org/z/zvbjvqf7n

kevmo314:
> Seems like the go compiler already performs this optimization for divisors that are powers of 2?
> 
> [godbolt.org/z/zvbjvqf7n](https://godbolt.org/z/zvbjvqf7n)

Yeah, I wouldn't be surprised -- the go compiler is quite smart. I don't think the benchmark results indicate there's an obvious improvement here, although it's worth noting that because `pageFileBytes` doesn't have to be a multiple of two that go might still not optimize it.

PR #198, opened 2024-03-25
feat: pass width upon index creation

friendlymatthew:
We were doing double work where in `handleJsonlObject`, we'd recompute the width value upon json token. Instead, we read the width directly from the metaslot

### DIFF
--- a/pkg/appendable/index_file.go
+++ b/pkg/appendable/index_file.go
@@ -154,38 +154,38 @@ func (i *IndexFile) IndexFieldNames() ([]string, error) {
 	return fieldNames, nil
 }
 
-func (i *IndexFile) FindOrCreateIndex(name string, fieldType FieldType) (*metapage.LinkedMetaSlot, error) {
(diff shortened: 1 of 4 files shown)


pkg/appendable/index_file.go
```diff
@@ -154,48 +154,48 @@ func (i *IndexFile) IndexFieldNames() ([]string, error) {
 	return fieldNames, nil
 }
 
-func (i *IndexFile) FindOrCreateIndex(name string, fieldType FieldType) (*metapage.LinkedMetaSlot, error) {
+func (i *IndexFile) FindOrCreateIndex(name string, fieldType FieldType) (*metapage.LinkedMetaSlot, uint16, error) {
```
kevmo314, line 157: I'd return the entire index meta instead.

```suggestion
func (i *IndexFile) FindOrCreateIndex(name string, fieldType FieldType) (*metapage.LinkedMetaSlot, *IndexMeta, error) {
```

kevmo314 (approved):

pkg/appendable/index_file.go
friendlymatthew: Yeah that's much more elegant. Thanks

PR #199, opened 2024-03-25
feat: define trigrams

### LINKED ISSUE #196: Create a new index type to represent full text search.

friendlymatthew:
Closes #196


pkg/handlers/jsonl.go
```diff
@@ -147,112 +148,135 @@ func (j JSONLHandler) handleJSONLObject(f *appendable.IndexFile, r []byte, dec *
...
-			case json.Number, float64:
-				buf := make([]byte, 8)
+
 				switch value := value.(type) {
-				case json.Number:
-					f, err := value.Float64()
-					if err != nil {
-						return fmt.Errorf("failed to parse float: %w", err)
+				case string:
+					if ft == appendable.FieldTypeString {
```
kevmo314, line 167: Can this be determined by the meta page?

friendlymatthew: Yeah, I was unsure if this was the best approach, so I decided to not touch it.

`jsonTypeToFieldType` returns a list of valid `FieldTypes` given the `value.(type)` If `value` is a `string`, it causes a variance in behavior. So when we loop by the different `fieldType`, I figured I'd add some guard to the logic below.

This leads to a lot of tests breaking which should occur since every string var should cause two indexes to form, but should I case by the fieldType instead of `value.(type)`?

kevmo314: Yeah it makes more sense to case on field type. Basically any type introspection at the handler level is left over from back when we had trees that stored multiple types. It's no longer necessary and in retrospect that was a bad idea so it's good we got rid of it :)

friendlymatthew: Sounds good. Note, we'll still have to do some `value` type introspection. Something like:

```go
switch ft {
     case FieldTypeTrigram:
          switch (value).type {
                  string: // do stuff
                  default: return fmt.Errorf("wrong type")
          }
}
```

Or is this roundabout, and we can directly cast to a string since we made the assertions above?

kevmo314: We can directly cast to a string, something like:

```go
switch ft {
  case FieldTypeTrigram:
    value, ok := value.(string)
    if !ok {
      return fmt.Errorf("expected a string didn't get a string")
    }
}
```

pkg/handlers/jsonl.go
```diff
@@ -147,112 +148,139 @@ func (j JSONLHandler) handleJSONLObject(f *appendable.IndexFile, r []byte, dec *
...
 					}
-				case json.Delim('{'):
-					// find the index to set the field type to unknown.
-					if err := j.handleJSONLObject(f, r, dec, append(path, key), data); err != nil {
-						return fmt.Errorf("failed to handle object: %w", err)
-					}
-					// read the }
-					if t, err := dec.Token(); err != nil || t != json.Delim('}') {
-						return fmt.Errorf("expected '}', got '%v'", t)
+				case appendable.FieldTypeArray, appendable.FieldTypeObject:
```
friendlymatthew, line 242: since these are degenerate cases (*as of now*), I hope it's alright to case them together and combine the logic

kevmo314: Yep that's fine.

pkg/trigram/trigram.go
```diff
@@ -9,19 +12,38 @@ type Trigram struct {
...
+			b.WriteRune(r)
+
+			fmt.Printf("valid: %v\n", i)
+		} else {
+			fmt.Printf("invalid: %v\n", i)
+		}
+	}
+	return b.String(), ogOffsets
+}
+
```
friendlymatthew, line 32: I've been picking my hair for the past hour trying to come up with an elegant solution. 

`normalizeToAscii` should normalize the string while returning a mapping of the original offers to the new index. 


this is because given `"h∫owd*y)__do "`, it normalizes it to `"howdydo "`, which is great. but we lose the offsets.


The problem right now is that since we're iterating through the unicode, the `i` doesn't match up. a sample output looks like:
```
valid: 0
invalid: 1
valid: 4
valid: 5
valid: 6
invalid: 7
valid: 10
```


Since invalid characters take more than one element.

pkg/trigram/trigram_test.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+func TestTrigram(t *testing.T) {
+	t.Run("test basic trigram", func(t *testing.T) {
+		p := "howdydo"
+
+		expected := [5]Trigram{
+			{
+				Word:   "how",
+				Offset: 0,
```
kevmo314, line 16: It seems like it might be more useful to store the byte range instead of the byte offset, is the byte offset sufficient to create a `DataPointer`?

pkg/trigram/trigram.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	return b.String(), ogOffsets
+}
+
+func Shuffle(trigrams []Trigram) []Trigram {
+	soup := make([]Trigram, len(trigrams))
+	copy(soup, trigrams)
+
+	rand.Seed(time.Now().UnixNano())
```
kevmo314, line 53: Can we seed the randomness based on the hash of a string? https://stackoverflow.com/a/13582881/86433

This way, the same string will always shuffle to the same result making it deterministically random.

kevmo314 (approved):

pkg/trigram/trigram.go
friendlymatthew: Sure but more importantly, I assume this is what we want for the typescript version

pkg/trigram/trigram_test.go
friendlymatthew: This is the current scheme for inserting an element:

```go
for _, tri := range trigrams {
     valueBytes := []byte(tri.Word)
     mp.Offset += tri.Offset
     mp.Length += uint32(tri.Offset)
     if err := page.BPTree(&btree.BPTree{Data: r, DataParser: j, Width: width}).Insert(
          btree.ReferencedValue{
               DataPointer: mp,
               Value:       valueBytes,
          }, data); err != nil {
          return fmt.Errorf("failed to insert into b+tree: %w", err)
     }
}
```

Since the existing `mp.Offset` is ` data.Offset + uint64(fieldOffset)`, it makes sense to just add the trigram offset to the existing offset.

I do that with length as well, but I can see why we'd want to store the `Length` as well.

Since given `ho¥wdy` will create the following trigrams:

```typescript
[{ word: "how", offset: 0, length: 4 (this is because `ho¥w`}, ...]
```

PR #200, opened 2024-03-28
feat: remove N offsets store

friendlymatthew:
The hex editor points to unexpected variance in behavior. Since we now store multi slots anyway, this PR removes the unnecessary N offset store.


src/file/index-file.ts
```diff
@@ -132,15 +132,24 @@ export class IndexFileV1<T> implements VersionedIndexFile<T> {
...
+        break;
+      }
+
+      const nextPage = new PageFile(this.resolver, nextOffset);
+      const nextMps = await nextPage.splitPage();
+      mps = [...mps, ...nextMps];
+
+      currMps = nextMps;
     }
+
```
friendlymatthew, line 152: we also had to iterate as we fetched the mps

kevmo314 (approved):
Wow nice work figuring out the cause

PR #202, opened 2024-03-29
feat: handle no results yielded

friendlymatthew:
The `TraversalRecord`'s first element returns a leaf node. When there are no fitting records, we return an index, `i`, right outside the bounds of `leafPointers`. That is from `0..len(leafPointers)-1`, index is usually `len(leafPointers)`. 

This PR adds one extra catch statement within the `init()` that checks this assertion. If true, we return false.

### DIFF
--- a/src/btree/traversal.ts
+++ b/src/btree/traversal.ts
@@ -29,11 +29,7 @@ export class TraversalIterator {
     const offset = rootResponse.pointer;
     this.records = await this.tree.traverse(this.key, root, offset);
 
-    if (this.records[0].index === this.records[0].node.leafPointers.length) {
-      return false;
-    }
-
-    return true;
+    return this.records[0].index !== this.records[0].node.numPointers();
   }
 
   getKey(): ReferencedValue {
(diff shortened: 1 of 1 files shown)


kevmo314:
Is the latest version of the code pushed? I'm not following how the commit matches the description

friendlymatthew:
> Is the latest version of the code pushed? I'm not following how the commit matches the description

Maybe I was being too clever. I figured we could inline the two statements, but then it's probably better to be explicit with the logic. I've restaged the logic to be more explicit

src/btree/traversal.ts
```diff
@@ -28,6 +28,11 @@ export class TraversalIterator {
     const root = rootResponse.rootNode;
     const offset = rootResponse.pointer;
     this.records = await this.tree.traverse(this.key, root, offset);
```
friendlymatthew, line 30: instead of directly updating the recursive `traverse()` logic, we can directly catch here

kevmo314:
> > Is the latest version of the code pushed? I'm not following how the commit matches the description
> 
> Maybe I was being too clever. I figured we could inline the two statements, but then it's probably better to be explicit with the logic. I've restaged the logic to be more explicit

Ahhh ok, nvm this was my bad I misread the diff, I thought the change was to make it `return true;` instead 🙃

Looks good, nice job finding that. I think the one-liner you had in the first commit is better, I just read the colors wrong for some reason.

kevmo314 (approved):

PR #204, opened 2024-04-01
fix: starting beyond tree causes error

kevmo314:
https://github.com/kevmo314/appendable/pull/202/files for golang

### DIFF
--- a/pkg/btree/bptree.go
+++ b/pkg/btree/bptree.go
@@ -80,7 +80,7 @@ func (p *TraversalIterator) init() bool {
 	}
 	p.records = path
 
-	return p.records[0].index != p.records[0].node.NumPointers()
+	return true
 }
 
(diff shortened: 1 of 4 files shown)


friendlymatthew:
So I've been stepping through this, and weirdly enough this assertion in the golang:

```go
return p.records[0].index != p.records[0].node.NumPointers()
```

gets returned as false when we start backwards. That is:
```go
iter, err := tree.Iter(ReferencedValue{Value: []byte{1, 2, 3, 4, 5, 6, 7, 8}, DataPointer: pointer.MemoryPointer{Offset: math.MaxUint64}})
		
for ; iter.Prev(); i-- {
    // never gets hit
}
```

friendlymatthew (approved):
I've resolved the tests and made some adjustments to the iterator

PR #210, opened 2024-04-03
feat: add coverage, rename conventions

friendlymatthew:
After changing the meta pages to be a slot rather than one page, the naming conventions weren't changed which was misleading.

This PR renames the test variables, as well as adding a test case that does the following.

Given N slots (30), we create a linked list. 
The # of slots should be 30, but the # of pages we get should be `N // 2 + 1`.


pkg/metapage/slot.go
```diff
@@ -87,11 +87,11 @@ func (m *LinkedMetaSlot) Reset() error {
...
-// Collect returns a slice of all linked meta pages from this page to the end.
+// Collect returns a slice of all linked meta slots from this page to the end.
 // This function is useful for debugging and testing, however generally it should
 // not be used for functional code.
 func (m *LinkedMetaSlot) Collect() ([]*LinkedMetaSlot, error) {
-	var pages []*LinkedMetaSlot
+	var slots []*LinkedMetaSlot
 	node := m
 	for {
 		exists, err := node.Exists()
```
friendlymatthew, line 97: so it's correctly adding the slots in `AddNext()`, however when we `Collect()`, it's an infinite loop because this isn't catching.

PR #211, opened 2024-04-04
fix: slot reset and exists

friendlymatthew:
`Reset()` and `Exists()` performed operations that would either wipe a whole page or check a page. Since we moved to slots, when we call these functions from the slot level, it should perform these operations at that slot size.

### DIFF
--- a/pkg/metapage/multi.go
+++ b/pkg/metapage/multi.go
@@ -86,18 +86,15 @@ func (m *MultiPager) Next(offset uint64) (*LinkedMetaSlot, error) {
 	return &LinkedMetaSlot{offset: next.Offset, pager: m}, nil
 }
 
-func (m *MultiPager) GetNextSlot(buf []byte) (int64, error) {
-	if buf != nil && len(buf) > m.rws.SlotSize() {
-		return 0, errors.New("buffer is too large")
-	}
+func (m *MultiPager) GetNextSlot() (int64, error) {
 
 	// find next available page offset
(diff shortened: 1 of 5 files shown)


pkg/metapage/multi.go
```diff
@@ -163,9 +163,17 @@ func (m *MultiPager) Exists(offset uint64) (bool, error) {
...
+}
+
+func (m *MultiPager) PageExists(offset uint64) (bool, error) {
+	return m.spaceExists(offset, m.rws.PageSize())
+}
+
+func (m *MultiPager) spaceReset(offset uint64, size int) error {
+	// write a full slot of zeros
+	emptyPage := make([]byte, size)
 	binary.LittleEndian.PutUint64(emptyPage[12:20], ^uint64(0))
```
kevmo314, line 177: Do slots actually need a pointer to the next slot? Or could they just do `offset + 256`?

kevmo314 (approved):

pkg/metapage/multi.go
friendlymatthew: Ah, we don't need a pointer to the next slot. The typescript version ingests a page, splits it into 16 slots, then uses the last slot's next pointer to fetch the next batch of slots.

I guess a possible refactor would be to remove all next pointers, and store them at the page level? I figured for consistency's sake, every slot contains the next pointer info

kevmo314: Yeah, I think it can be stored at the page level. Although it doesn't really affect the storage capacity all things considered.

PR #217, opened 2024-04-05
feat: store unique pointers for trigrams

friendlymatthew:
When we handle a JSONL object line, we store a memory pointer of the following:

```go
mp := pointer.MemoryPointer{
    Offset: data.Offset + uint64(fieldOffset),
    Length: uint32(dec.InputOffset() - fieldOffset),
}
```

For the trigram case, we split a string into a bunch of trigrams that conveniently store the offset and the holistic length. We need to make sure when we insert the trigram into the btree, we update the memory pointer correctly.

The PR changes the faulty logic, which would iteratively increment the memory pointer's offset + length. However, this doesn't make sense as the offset for that iteration should be added to the offset from the original memory pointer.

### DIFF
--- a/pkg/handlers/jsonl.go
+++ b/pkg/handlers/jsonl.go
@@ -187,10 +187,16 @@ func (j JSONLHandler) handleJSONLObject(f *appendable.IndexFile, r []byte, dec *
 
 					for _, tri := range trigrams {
 						valueBytes := []byte(tri.Word)
-						mp.Offset += tri.Offset
+
+						newMp := pointer.MemoryPointer{
+							Offset: mp.Offset + tri.Offset,
+							Length: mp.Length + tri.Length,
+						}
+
 						mp.Length += tri.Length
+
 						if err := page.BPTree(&btree.BPTree{Data: r, DataParser: j, Width: width}).Insert(btree.ReferencedValue{
-							DataPointer: mp,
+							DataPointer: newMp,
 							Value:       valueBytes,
 						}, data); err != nil {
 							return fmt.Errorf("failed to insert into b+tree: %w", err)
--- a/pkg/handlers/jsonl.go
+++ b/pkg/handlers/jsonl.go
@@ -188,14 +188,12 @@ func (j JSONLHandler) handleJSONLObject(f *appendable.IndexFile, r []byte, dec *
 					for _, tri := range trigrams {
 						valueBytes := []byte(tri.Word)
 
-						newMp := pointer.MemoryPointer{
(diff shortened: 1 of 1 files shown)


pkg/handlers/jsonl.go
```diff
@@ -187,10 +187,16 @@ func (j JSONLHandler) handleJSONLObject(f *appendable.IndexFile, r []byte, dec *
...
 						valueBytes := []byte(tri.Word)
-						mp.Offset += tri.Offset
+
+						newMp := pointer.MemoryPointer{
+							Offset: mp.Offset + tri.Offset,
+							Length: mp.Length + tri.Length,
+						}
+
 						mp.Length += tri.Length
+
```
friendlymatthew, line 197: Just want to make sure this is the correct way to store the length?

For every iteration, we should keep track of the length + iteratively sum the lengths?

kevmo314: I don't think this is correct. The offset should be added, but the length only needs to represent the length of the trigram, so it would be something like `DataPointer: pointer.MemoryPointer{ Offset: <the two added>, Length: tri.Length }`?

friendlymatthew: Thank you -- this makes sense.

pkg/handlers/jsonl.go
```diff
@@ -187,10 +187,14 @@ func (j JSONLHandler) handleJSONLObject(f *appendable.IndexFile, r []byte, dec *
 
 					for _, tri := range trigrams {
 						valueBytes := []byte(tri.Word)
-						mp.Offset += tri.Offset
-						mp.Length += tri.Length
+
+						newMp := pointer.MemoryPointer{
+							Offset: mp.Offset + tri.Offset,
+							Length: tri.Length,
```
friendlymatthew, line 193: should be good now. Indexes great with the palo-alto dataset

pkg/handlers/jsonl.go
```diff
@@ -187,10 +187,14 @@ func (j JSONLHandler) handleJSONLObject(f *appendable.IndexFile, r []byte, dec *
...
+						}
+
 						if err := page.BPTree(&btree.BPTree{Data: r, DataParser: j, Width: width}).Insert(btree.ReferencedValue{
-							DataPointer: mp,
+							DataPointer: newMp,
```
kevmo314, line 197: I'd inline the struct, no need to pull it away from the definition.

kevmo314 (approved):

PR #220, opened 2024-04-08
feat: ngrams

friendlymatthew:
This PR contains a ngram tokenizer that can produce tokens of varying length N, where N can be 1, 2, or 3.

*Note: to avoid accounting for variable width, the fixed with for a FieldTypeNgram is 3. We prepend any shorter tokens with whitespace*

Elastic Search uses both a trigram + 12gram (min gram: 1, max gram: 2) for searching. Would be good to see and compare


pkg/appendable/appendable.go
```diff
@@ -162,8 +162,8 @@ func DetermineType(ft FieldType) uint16 {
 		width = uint16(shift + 0)
 	case FieldTypeFloat64, FieldTypeInt64, FieldTypeUint64:
 		width = uint16(shift + 8)
-	case FieldTypeTrigram:
-		width = uint16(shift + trigram.N)
+	case FieldTypeNgram:
+		width = uint16(shift + ngram.MAX_GRAM) // we'll bound all ngrams with the max len which is N = 3
```
kevmo314, line 166: Is this storing different n-gram lengths in the same b-tree? I don't think we should do that because the disk space cost of storing the lengths will be very high.

If we want to store multiple length grams, we can do it by creating a separate index. The main observation is that it never makes sense to compare across n-gram lengths, like comparing a 2-gram against a 3-gram has no semantic meaning and n-grams should only be compared across the same length. Thus, making a separate index is probably more efficient.

friendlymatthew: Yeah this makes total sense. Will make two additional indexes (1-gram and 2-gram)

pkg/appendable/appendable.go
```diff
@@ -162,8 +163,12 @@ func DetermineType(ft FieldType) uint16 {
...
 	case FieldTypeFloat64, FieldTypeInt64, FieldTypeUint64:
 		width = uint16(shift + 8)
-	case FieldTypeTrigram:
-		width = uint16(shift + trigram.N)
+	case FieldType3gram:
+		width = uint16(shift + 3)
+	case FieldType2gram:
+		width = uint16(shift + 2)
+	case FieldType1gram:
+		width = uint16(shift + 1)
```
friendlymatthew, line 171: each now have a fixed width for it's respective gram length

pkg/appendable/appendable.go
```diff
@@ -54,12 +53,14 @@ const (
 	FieldTypeArray
 	FieldTypeBoolean
 	FieldTypeNull
-	FieldTypeTrigram
+	FieldType3gram
```
kevmo314, line 56: Nitpick but this would be `FieldTypeTrigram`, `FieldTypeBigram`, and `FieldTypeUnigram`.

kevmo314 (approved):

PR #221, opened 2024-04-08
feat: ngram tokenizer

friendlymatthew:
Tokenizer on the typescript side.

To initialize, you need a `min_gram` and a `max_gram`. Also updated the search DSL params to include gram ranges.

### DIFF
--- a/src/db/database.ts
+++ b/src/db/database.ts
@@ -396,9 +396,9 @@ export class Database<T extends Schema> {
   search(
     key: keyof T,
     like: string,
-    config: { minGram: number; maxGram: number },
+    config?: { minGram: number; maxGram: number },
   ) {
-    let { minGram, maxGram } = config;
+    let { minGram = 1, maxGram = 2 } = config || {};
 
     const search: Search<T> = {
       key,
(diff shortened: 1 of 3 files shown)


src/db/database.ts
```diff
@@ -390,10 +393,18 @@ export class Database<T extends Schema> {
     return new QueryBuilder(this).where(key, operation, value);
   }
 
-  search(key: keyof T, like: string) {
+  search(
+    key: keyof T,
+    like: string,
+    config: { minGram: number; maxGram: number },
```
kevmo314, line 399: Make this optional and add a default

src/search/tokenizer.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  constructor(minGram: number, maxGram: number) {
+    this.maxGram = maxGram;
+    this.minGram = minGram;
+  }
+
+  tokens(phrase: string): string[] {
+    let ngrams: string[] = [];
+
+    let wordOffsets = [];
```
kevmo314, line 13: Add a type annotation

kevmo314 (approved):

PR #223, opened 2024-04-08
feat: update search flow

friendlymatthew:
Sorry for the large PR. 

## What this PR contains:

- introduce `Unigram` and `Bigram` to ts `FieldType`s
- add `NgramToken` type that encodes the value plus the token gram type
- Update the search flow
Given a phrase to search for, we break it into ngram tokens, shuffle it, then we iterate through. For every ngram type, we keep a btree cache to remove the need for repeating logic. We append the result to the NgramTable, which acts as a "priority queue" and yields the most frequent result

- revert #222 as this is not planned anymore
- a test util function to simplify text encoding -> buffer


friendlymatthew:
I will split this into various PRs.

PR #225, opened 2024-04-08
feat: add ngram token type

friendlymatthew:
This PR introduces the `NgramToken` type and updates the logic in Tokenizer.

Added a static `TextEncoder` to `Tokenizer()` which will encode each trigram's strings to an ArrayBuffer. It also contains the  ngram token's specific FieldType.Ngram type.

### DIFF
--- a/src/search/tokenizer.ts
+++ b/src/search/tokenizer.ts
@@ -1,14 +1,30 @@
+import { FieldType } from "../db/database";
+
+export type NgramToken = {
+  value: string;
+  valueBuf: ArrayBuffer;
+  type: FieldType;
+};
+
 export class NgramTokenizer {
   private readonly minGram: number;
   private readonly maxGram: number;
+
+  private allGrams: Map<number, FieldType> = new Map<number, FieldType>([
+    [1, FieldType.Unigram],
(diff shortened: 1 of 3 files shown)


src/search/tokenizer.ts
```diff
@@ -29,6 +45,12 @@ export class NgramTokenizer {
     }
 
     for (let N = this.minGram; N <= this.maxGram; N++) {
+      let gType =
+        this.allGrams.get(N) ??
+        (() => {
+          throw new Error(`Unrecognized gram type for gram length: ${N}`);
+        })();
```
kevmo314, line 52:
```suggestion
     if (!gType) {
          throw new Error(`Unrecognized gram type for gram length: ${N}`);
     }
```

src/search/tokenizer.ts
```diff
@@ -29,6 +45,12 @@ export class NgramTokenizer {
     }
 
     for (let N = this.minGram; N <= this.maxGram; N++) {
+      let gType =
```
kevmo314, line 48:
```suggestion
      const gType =
```

kevmo314 (approved):

PR #226, opened 2024-04-08
feat: update search flow

friendlymatthew:
## Search Flow
Given a term to search for, `like`, tokenize then shuffle the tokens. As we iterate through, we keep a bptree cache for better lookups. We iterate through, and for every ngram match, we insert into the `NgramTable`


src/db/database.ts
```diff
@@ -114,31 +114,42 @@ export class Database<T extends Schema> {
...
+      const likeToks = NgramTokenizer.shuffle(tok.tokens(like));
+
+      const ngramTable = new NgramTable();
+      const bptreeCache = new Map<FieldType, BPTree>();
+
+      for (const token of likeToks) {
+        const { type: fieldType, valueBuf } = token;
+        let bptree = bptreeCache.get(fieldType);
+
+        if (!bptree) {
```
friendlymatthew, line 128: Since search supports ngrams, a btree cache was needed to efficiently iterate through the respective ngram tree.

kevmo314: Curious, why? Isn't the page-level cache sufficient?

friendlymatthew: The tokens produced have various ranges, that is for a 12gram, the tokens will either reside in the `Unigram` or `Bigram` page. Upon every token, we can call `seek()` and `new BPTree()`, but I figured this would be more efficient

kevmo314: Ahh, I see because the index meta itself isn't cached?

I would opt to cache the index meta pages instead of the entire btree object. It is slightly more expensive but the btree object is ideally stateless so it's not that problematic to reinstantiate.

That being said, we do cache individual pages right now, right?

friendlymatthew: Yes we cache meta pages in `index-file.ts`: https://github.com/kevmo314/appendable/blob/main/src/file/index-file.ts#L130

friendlymatthew: > I would opt to cache the index meta pages instead of the entire btree object. It is slightly more expensive but the btree object is ideally stateless so it's not that problematic to reinstantiate.

Will do! FWIW, this isn't a traditional "cache" as there is no network request. Given an in-memory list of meta pages, this hashmap aims to reduce the time complexity needed to traverse this meta page list for the correct bptree. We do this because there can only be three bptree variants.

kevmo314 (approved):

PR #232, opened 2024-04-09
Mar/Unmarshall node

### LINKED ISSUE #189: Implement repeated DataPointers
For some values it's unnecessary to store the values of adjacent DataPointers. For example, if we have 10 key "abcd"'s, we don't need to store "abcd" 10 times, we can store it once and write a byte that indicates that the next keys are identical.

Note: this is a little tricky when splitting nodes as we cannot cross node boundaries. In that light, it would be best to make this change in `MarshalBinary`/`UnmarshalBinary` and let the nodes store multiple copies in the actual memory representation. Essentially, in `MarshalBinary`, add a flag bit (maybe the first bit of the length? not sure what to do for fixed width) that indicates if it's a repeated value. If it is, skip encoding the value. In `UnmarshalBinary`, do the opposite.

friendlymatthew:
Fix #189

### DIFF
--- a/pkg/btree/bptree_test.go
+++ b/pkg/btree/bptree_test.go
@@ -715,17 +715,26 @@ func TestBPTree_Iteration_StartsAfterTree(t *testing.T) {
 
 	for i := 0; i < count; i++ {
 		buf := []byte{0x01}
-		if err := tree.Insert(ReferencedValue{Value: buf, DataPointer: pointer.MemoryPointer{Offset: uint64(i)}}, pointer.MemoryPointer{Offset: uint64(i), Length: uint32(len(buf))}); err != nil {
+		if err := tree.Insert(
+			ReferencedValue{
+				Value: buf,
+				DataPointer: pointer.MemoryPointer{
+					Offset: uint64(i),
+					Length: uint32(2),
+				},
+			},
+			pointer.MemoryPointer{
+				Offset: uint64(i),
+				Length: uint32(len(buf)),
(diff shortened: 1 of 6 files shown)


pkg/btree/bptree_test.go
```diff
@@ -614,7 +614,19 @@ func TestBPTree_Iteration_Overcount(t *testing.T) {
 		buf := make([]byte, 8)
 		binary.BigEndian.PutUint64(buf, math.Float64bits(23))
 
-		if err := tree.Insert(ReferencedValue{Value: buf, DataPointer: pointer.MemoryPointer{Offset: uint64(i)}}, pointer.MemoryPointer{Offset: uint64(i), Length: uint32(len(buf))}); err != nil {
+		if err := tree.Insert(
+			ReferencedValue{
+				Value: buf,
+				DataPointer: pointer.MemoryPointer{
+					Offset: uint64(i),
+					Length: uint32(9),
```
friendlymatthew, line 622: Some test cases failed because of one use case where we attempted to insert multiple items, but we zeroed out the MemoryPointer. This is an issue because for identical nodes, we negate `Length` to indicate there should be a copy. However, `-0` is `0` so it will attempt to read from the buffer which is not correct.

kevmo314: Ahhh, to address that, store `-(length + 1)` instead. Conveniently, this is just the `~` operator, so store `~length` instead of `-length`.

PR #234, opened 2024-04-10
feat: add cursor

friendlymatthew:
A Btree cursor traverses the btree and returns high level information about the btree. 

The difference between this class and the traversal iterator, is that the iterator is stateful, while the Cursor takes a "static" view of the instantiated btree

### DIFF
--- a/src/btree/cursor.ts
+++ b/src/btree/cursor.ts
@@ -1,0 +1,81 @@
+import { BTree, ReferencedValue } from "./btree";
+import { BTreeNode } from "./node";
+
+export class BTreeCursor {
+  private readonly btree: BTree;
+
+  private uniqueEntriesPromise: Promise<number> | null = null;
+
+  constructor(btree: BTree) {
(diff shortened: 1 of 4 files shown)


src/btree/cursor.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  // Counts the number of unique entries in the datafile
+  uniqueEntries(): Promise<number> {
+    if (!this.uniqueEntriesPromise) {
+      this.uniqueEntriesPromise = this.computeUniqueLeafs();
+    }
+
+    return this.uniqueEntriesPromise;
+  }
```
friendlymatthew, line 54: This is needed to compute `N` in the ranking fn: 

```
N is the total number of documents on the content node.
```

In this case, this would be size of a set containing all leaf nodes' memory pointer offset

friendlymatthew:
Closing via #235 and a cursor is not idiomatic in ts

PR #237, opened 2024-04-11
feat: Add TotalLength to Index Meta

friendlymatthew:
This PR adds an extra attribute to the `IndexMeta` called `TotalLength`. For a given metapage, `TotalLength` is the trailing sum of the length of every data entry.


pkg/appendable/appendable.go
```diff
@@ -134,17 +134,19 @@ func (m *FileMeta) UnmarshalBinary(buf []byte) error {
 }
 
 type IndexMeta struct {
-	FieldName string
-	FieldType FieldType
-	Width     uint16
+	FieldName   string
+	FieldType   FieldType
+	Width       uint16
+	TotalLength uint64
```
kevmo314, line 140: Maybe name this something like `TotalValueLength`? It's not clear what length it's referring to. A comment would also be good.

kevmo314 (approved):

PR #243, opened 2024-04-15
chore: remove n offsets code

kevmo314:
Looks like this was still left in, bad merge maybe?


friendlymatthew:
Oh, since we reverted back to the meta page layout, we needed to store the next n offsets so that we could fetch efficiently:

https://github.com/kevmo314/appendable/blob/main/src/file/index-file.ts#L104

PR #253, opened 2024-04-18
chore: add debug assertion to prevent read/write across page boundaries

kevmo314:


friendlymatthew (approved):
So awesome. Thank you!

PR #257, opened 2024-04-18
feat: change metadata

friendlymatthew:
Since we allocate 1 byte for the length, we should seek by `1+`, not `4+`


friendlymatthew:
Weirdly, the previous version snuck through our testing... I will look into coverage for this

friendlymatthew:
Closing since #258 won

PR #258, opened 2024-04-18
Marshal coverage

friendlymatthew:
Some test cases for `appendable.go` like creating a new tree, adding some pages, marshaling the metadata

### DIFF
--- a/pkg/metapage/multi.go
+++ b/pkg/metapage/multi.go
@@ -105,7 +105,7 @@ func (m *LinkedMetaPage) Metadata() ([]byte, error) {
 	if _, err := m.rws.Seek(int64(m.rootMemoryPointerPageOffset())+24, io.SeekStart); err != nil {
 		return nil, err
 	}
-	buf := make([]byte, 4+m.rws.SlotSize())
+	buf := make([]byte, 1+m.rws.SlotSize())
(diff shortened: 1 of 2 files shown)


pkg/metapage/multi.go
```diff
@@ -105,7 +105,7 @@ func (m *LinkedMetaPage) Metadata() ([]byte, error) {
 	if _, err := m.rws.Seek(int64(m.rootMemoryPointerPageOffset())+24, io.SeekStart); err != nil {
 		return nil, err
 	}
-	buf := make([]byte, 4+m.rws.SlotSize())
+	buf := make([]byte, 1+m.rws.SlotSize())
```
friendlymatthew, line 108: The test cases above passes regardless of what this number is set to. Do you have a suggestion for a breaking test case? A bit confused why we used to pad it with 4

kevmo314: Ahh this is only the read buffer. Actually a value of `4` here would be fine, it will just overallocate the buffer by a bit. It will get truncated on line 114.

kevmo314: In other words, I don't think you can create a test case that actually checks this.

kevmo314 (approved):

PR #263, opened 2024-04-23
fix: add extra assertion to test

friendlymatthew:


friendlymatthew (commented):
Still investigating why the index file doesn't capture the rest of the field headers. I think the fault lies within the `AddNext()` logic.

PR #268, opened 2024-04-23
feat: no zero ranges

friendlymatthew:


src/resolver/range-request.ts
```diff
@@ -39,6 +39,14 @@ export async function requestRanges(
     return await resolveIndividualPromises(url, ranges);
   }
 
+  for (const { start, end } of ranges) {
+    if (end - start === 0) {
```
kevmo314, line 43:
```suggestion
    if (end - start <= 0) {
```

Negative would also be bad.

kevmo314 (approved):

PR #273, opened 2024-04-25
feat: remove buffer

friendlymatthew:
Remove buffer, use promise

### DIFF
--- a/src/file/multi.ts
+++ b/src/file/multi.ts
@@ -9,14 +9,13 @@ const LENGTH_BYTES = 4;
 const COUNT_BYTE = 1;
 
 export class LinkedMetaPage {
-  private metaPageDataPromise?: Promise<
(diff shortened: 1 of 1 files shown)


src/file/multi.ts
```diff
@@ -85,12 +80,7 @@ export class LinkedMetaPage {
     const count = view.getUint8(POINTER_BYTES);
 
     if (this.index < count - 1) {
-      return new LinkedMetaPage(
-        this.resolver,
-        this.offset,
-        this.index + 1,
-        this.metaPageData,
-      );
+      return new LinkedMetaPage(this.resolver, this.offset, this.index + 1);
```
kevmo314, line 83: The promise still needs to be passed here, right? Otherwise the cache won't persist across two slots' `LinkedMetaPage`s.

friendlymatthew: Done. Thank you

PR #277, opened 2024-04-30
Eucqueue

friendlymatthew:
This PR has a EucQueue which is an abstraction of `heap` and makes a `minQueue` or `maxQueue`. 

The items in the queue are prioritized by the euclidian distance of a given position and that item's position

### DIFF
--- a/pkg/hnsw/eucqueue_test.go
+++ b/pkg/hnsw/eucqueue_test.go
@@ -1,0 +1,61 @@
+package hnsw
+
+import (
+	"testing"
+)
+
+type Pair struct {
+	id   NodeID
+	dist float64
+}
+
+func TestEucQueue(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pkg/hnsw/eucqueue_test.go
```diff
@@ -0,0 +1,61 @@
...
+		i := 0
+		for eq.Len() > 0 {
+			item := eq.Pop()
+			if item.id != expected[i].id {
+				t.Fatalf("expected item %v, got %v at %v", expected[i].id, item.id, i)
+			}
+
+			i++
+		}
+	})
```
friendlymatthew, line 60: this is all a bit suspect. 

How should I approach this? I assume we still want out vector to be of type `[]float64`, but how do we deal with inaccuracies?

PR #280, opened 2024-05-01
feat: search layer

friendlymatthew:
This PR implements the `SearchLayer` algorithm.


pkg/hnsw/eucqueue.go
```diff
@@ -75,6 +79,13 @@ func (pq *MinQueue) Push(x any) {
 	*pq = append(*pq, item)
 }
 
+func (pq *MinQueue) Peek() *Item {
+	if len(*pq) == 0 {
+		return nil
+	}
+	return (*pq)[0]
+}
```
kevmo314, line 87: Since the base type is `[]*Item` and this operation doesn't mutate the array itself, it would be better to make this a value func:

```suggestion
func (pq MinQueue) Peek() *Item {
	if len(pq) == 0 {
		return nil
	}
	return pq[0]
}
```

pkg/hnsw/visited.go
```diff
@@ -0,0 +1,42 @@
+package hnsw
+
+import "fmt"
+
+type Set struct {
```
kevmo314, line 5: What's the benefit of this type over `map[NodeID]interface{}` directly?

```
x := map[NodeID]interface{}
len(x)
```

works to fetch the size.

pkg/hnsw/hnsw.go
```diff
@@ -59,3 +59,54 @@ func (h *Hnsw) EntryTopLayer() int {
 func (h *Hnsw) SpawnLayer() int {
 	return int(math.Floor(-math.Log(rand.Float64() * h.levelMultiplier)))
 }
+
+// w must be a max euc queue
+func (h *Hnsw) searchLayer(q Vector, ef, layerId int, w *EucQueue) {
```
kevmo314, line 64: return/handle the `err`s

kevmo314 (approved):

pkg/hnsw/visited.go
friendlymatthew: Yeah, I did this to maintain readability when I implemented the algorithm. I can remove this abstraction 👍

pkg/hnsw/hnsw.go
friendlymatthew: The only error raised was when computing the `Euclid Distance`. However, once merging the latest changes which refactors the euclid distance to only return the distance, we don't have any errors to propagate in `SearchLayer`.

PR #287, opened 2024-05-01
feat: knn search

friendlymatthew:
Screen Shot 2024-05-01 at 5 12 53 PM

### DIFF
--- a/pkg/hnsw/hnsw.go
+++ b/pkg/hnsw/hnsw.go
@@ -75,7 +75,7 @@ func (h *Hnsw) spawnLayer() int {
 searchLayer needs two things:
 1. todo! an item from a euc queue that computes the distance from the entry point node -> q.
 */
-func (h *Hnsw) searchLayer(q Vector, entryNode *Node, ef int, layerId int, numNearestToQ *MinQueue) {
+func (h *Hnsw) searchLayer(q Vector, entryNode *Node, ef int, layerId int) *MinQueue {
 
 	// visited is a bitset that keeps track of all nodes that have been visited.
 	// we know the size of visited will never exceed len(h.Nodes)
@@ -135,10 +135,13 @@ func (h *Hnsw) searchLayer(q Vector, entryNode *Node, ef int, layerId int, numNe
 		}
 	}
 
+	numNearestToQ := NewMinQueue()
 	for !nearestNeighborsToQForEf.IsEmpty() {
 		peeled := nearestNeighborsToQForEf.Peel()
 		numNearestToQ.Insert(peeled.id, peeled.dist)
 	}
+
+	return numNearestToQ
 }
 
 func (h *Hnsw) selectNeighbors(candidates *MinQueue, numNeighborsToReturn int) *MinQueue {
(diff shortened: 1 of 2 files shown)


pkg/hnsw/hnsw.go
```diff
@@ -74,17 +75,25 @@ func (h *Hnsw) spawnLayer() int {
 searchLayer needs two things:
 1. todo! an item from a euc queue that computes the distance from the entry point node -> q.
 */
-func (h *Hnsw) searchLayer(q Vector, ef, layerId int, nearestNeighborsToQForEf *MaxQueue) {
+func (h *Hnsw) searchLayer(q Vector, entryNode *Node, ef int, layerId int, numNearestToQ *MinQueue) {
```
friendlymatthew, line 78: @kevmo314 

`searchLayer` updates `numNearestToQ` BWO passing via params. Is this a good pattern?

kevmo314: It's ok but not ideal. Generally to avoid this, the `Hnsw` struct should own that queue.

pkg/hnsw/hnsw.go
```diff
@@ -145,3 +159,26 @@ func (h *Hnsw) selectNeighbors(candidates *MinQueue, numNeighborsToReturn int) *
...
+	if currentNearestElements.Len() < kNeighborsToReturn {
+		panic("")
+	}
+
+	kItems, err := currentNearestElements.Take(kNeighborsToReturn)
+	if err != nil {
+		return nil, err
+	}
+
+	return kItems, nil
```
kevmo314, line 183:
```suggestion
	return currentNearestElements.Take(kNeighborsToReturn)
```

pkg/hnsw/hnsw.go
friendlymatthew: > It's ok but not ideal. Generally to avoid this, the `Hnsw` struct should own that queue.

I'm a bit skeptical about making `Hnsw` own that queue, as these queues should be generated dynamically when we `Search` or `Insert`. I refactored the helper function to return a `MinQueue`, which I drain into the OG minqueue

pkg/hnsw/hnsw.go
```diff
@@ -145,3 +162,32 @@ func (h *Hnsw) selectNeighbors(candidates *MinQueue, numNeighborsToReturn int) *
...
+	currentNearestElements := NewMinQueue()
+	entryPointNode := h.Nodes[h.EntryNodeId]
+
+	for l := entryPointNode.layer; l >= 1; l-- {
+		numNearestToQAtLevelL := h.searchLayer(q, entryPointNode, 1, l)
+
+		for !numNearestToQAtLevelL.IsEmpty() {
+			peeled := numNearestToQAtLevelL.Peel()
+			currentNearestElements.Insert(peeled.id, peeled.dist)
+		}
```
friendlymatthew, line 176: for a given layer, get a minqueue of the nearest neighbors nearest to q. we have to add it to the currentNearestElements.

Is there a better way to do this? Maybe encapsulating this logic inside `queue` class

pkg/hnsw/hnsw.go
kevmo314: Yeah that approach works 👍

pkg/hnsw/hnsw.go
kevmo314: I think this is ok, probably some kind of bulk insert op would be better but it's not a huge deal because there might not be that much improvement with the heap.

PR #291, opened 2024-05-13
refactor select neighbor

friendlymatthew:
The select neighbor algorithm can be condensed to the following.

### DIFF
--- a/pkg/hnsw/hnsw_test.go
+++ b/pkg/hnsw/hnsw_test.go
@@ -1,7 +1,6 @@
 package hnsw
 
 import (
-	"fmt"
 	"testing"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/hnsw/hnsw_test.go
```diff
@@ -21,29 +21,53 @@ func TestHnswSelect(t *testing.T) {
...
 		if cn.Len() != 10 {
 			t.Fatalf("did not take 10 items")
 		}
 
+		expected := 11
+		i := 0
 		for !cn.IsEmpty() {
 			peeled := cn.Peel()
 			fmt.Printf("%v", peeled.id)
+			if peeled.id != NodeId(expected) {
```
friendlymatthew, line 53: this is where the assertion fails

PR #296, opened 2024-05-16
Inserting the first vector after the entry node

friendlymatthew:
A hnsw is guaranteed to have an entry node, that is a graph will never have zero vectors.

The insert `vector q` function iterates through the layers and finds the closest nearest neighbor. Note, the parameter explicitly states to find 1 neighbor.

The problem occurs in `searchLayer` line 114 where we iterate through candidates. The only candidate should be the entry node, but entry nodes don't have any friends at any levels!

Maybe I missed a part where an entry node should be friends with everyone, but I couldn't find that in the algorithm

### DIFF
--- a/pkg/hnsw/hnsw.go
+++ b/pkg/hnsw/hnsw.go
@@ -54,7 +54,7 @@ func NewHNSW(d, m int, efc int, entryPoint *Node) *Hnsw {
 		Nodes:           nt,
 		EntryNodeId:     enId,
 		NextNodeId:      nextId,
-		MaxLayer:        -1,
+		MaxLayer:        entryPoint.layer,
 		levelMultiplier: 1 / math.Log(float64(m)),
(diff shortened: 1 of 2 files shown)


pkg/hnsw/hnsw.go
```diff
@@ -109,9 +109,12 @@ func (h *Hnsw) searchLayer(q Vector, entryNode *Node, ef int, layerId int) (*Bas
 			break
 		}
 
+		fmt.Println("we should be checking")
+
 		friends := h.Nodes[closestCandidate.id].GetFriendsAtLevel(layerId)
```
friendlymatthew, line 114: specifically here, when we try to `GetFriendsAtLevel`, the entry node has no friends, as the graph has been freshly created

friendlymatthew:
Closing this PR and will reopen a new one

PR #298, opened 2024-05-17
feat: find correct entry node

friendlymatthew:
For `Insert` and `KNNSearch`, we need to traverse from the top layer .... qLayer (not included) and find the correct entry node. 

The correct entry node is the node who's max layer is greater than qLayer and is most closest to q amongst the entry node friends

### DIFF
--- a/pkg/hnsw/hnsw_test.go
+++ b/pkg/hnsw/hnsw_test.go
@@ -1,7 +1,6 @@
 package hnsw
 
 import (
-	"fmt"
 	"testing"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/hnsw/hnsw_test.go
```diff
@@ -266,3 +266,81 @@ func TestHnsw_Link(t *testing.T) {
...
+		q := []float64{6, 6}
+		qLayer := 3
+
+		// suppose we had m := []float{5, 5}. It is closer to q, so let's add m to the friends of ep
+		m := NewNode(1, []float64{5, 5}, 9)
+		h.Nodes[m.id] = m
+		mDist := m.VecDistFromVec(q)
+		for level := 9; level > qLayer; level-- {
+			fmt.Printf("level: %v", level)
+			h.Nodes[h.EntryNodeId].InsertFriendsAtLevel(level, m.id, mDist)
```
friendlymatthew, line 326: When iterating from the topmost level to the qLevel + 1, it seems the `InsertFriendsAtLevel` only updates the friend map at key: 9.

That is:
```
level: 10: friends: 0

level: 9: friends: 1

level: 8: friends: 0

level: 7: friends: 0

level: 6: friends: 0

level: 5: friends: 0

level: 4: friends: 0
```

PR #305, opened 2024-05-21
fix level generation bug

friendlymatthew:
Screenshot 2024-05-21 at 2 01 35 PM

The `spawnLayer` was incorrectly computed. Additional asserts to check leveling.

### DIFF
--- a/pkg/hnsw/hnsw.go
+++ b/pkg/hnsw/hnsw.go
@@ -53,7 +53,7 @@ func (h *Hnsw) getNextNodeId() NodeId {
 }
 
 func (h *Hnsw) spawnLevel() int {
-	return int(math.Floor(-math.Log(rand.Float64() * h.levelMultiplier)))
+	return int(math.Floor(-math.Log(rand.Float64()) * h.levelMultiplier))
 }
 
 func (h *Hnsw) searchLevel(q Vector, entryNodeItem *Item, ef int, levelId int) (*BaseQueue, error) {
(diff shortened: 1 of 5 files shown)


pkg/hnsw/hnsw.go
```diff
@@ -281,3 +281,35 @@ func (h *Hnsw) Insert(q Vector) error {
...
+	for k := range h.Nodes {
+		nodeIds = append(nodeIds, k)
+		num += 1
+	}
+
+	return nodeIds, num
+}
+
+// should never exceed MMAX, at level 0 shouldn't exceed MMAX0
+func (h *Hnsw) assert_num_neighbor_connections() (bool, error) {
```
kevmo314, line 300: Can you put this behind a compile guard to only run on debug builds?

pkg/hnsw/hnsw.go
```diff
@@ -281,3 +281,35 @@ func (h *Hnsw) Insert(q Vector) error {
...
+func (h *Hnsw) numNodes() ([]NodeId, int) {
+	num := 0
+
+	var nodeIds []NodeId
+	for k := range h.Nodes {
+		nodeIds = append(nodeIds, k)
+		num += 1
+	}
+
+	return nodeIds, num
```
kevmo314, line 296:
```suggestion
	return nodeIds, len(nodeIds)
```

pkg/hnsw/hnsw.go
```diff
@@ -281,3 +281,35 @@ func (h *Hnsw) Insert(q Vector) error {
 
 	return nil
 }
+
+// functions below are primarily used for testing
```
kevmo314, line 285: If they're only used for testing can they be in the `_test.go` file?

pkg/hnsw/hnsw_debug.go
```diff
@@ -0,0 +1,28 @@
+//go:build debug
+// +build debug
```
friendlymatthew, line 2: cmd: `go build -tags debug`

friendlymatthew:
Closing due to #307

PR #308, opened 2024-05-22
Update Level Initialization

friendlymatthew:
The entry node is initialized as 0, and as the # of nodes increase, the Hnsw graph automatically handles growing the layer.

That is, when creating a `NewHnsw()`, the entry node specified should always be initialized to 0.

### DIFF
--- a/pkg/hnsw/hnsw.go
+++ b/pkg/hnsw/hnsw.go
@@ -38,7 +38,6 @@ func NewHNSW(m int, efc int, entryPoint *Node) *Hnsw {
 		Nodes:           nt,
 		EntryNodeId:     enId,
 		NextNodeId:      enId + 1,
-		MaxLevel:        entryPoint.level,
 		levelMultiplier: 1 / math.Log(float64(m)),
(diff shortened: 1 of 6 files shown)


pkg/hnsw/hnsw.go
```diff
@@ -28,17 +28,18 @@ type Hnsw struct {
 	levelMultiplier float64
 }
 
-func NewHNSW(m int, efc int, entryPoint *Node) *Hnsw {
+func NewHNSW(m int, efc int, entryVector Vector) *Hnsw {
```
kevmo314, line 31: Do we check the dimension of the vector? We might want to add a dim parameter to the HNSW to ensure that all the vectors are the correct dimension to avoid a footgun.

kevmo314 (approved):

pkg/hnsw/hnsw.go
friendlymatthew: That was originally included but I took it out as I was verifying the correctness. 

Will include!

Issue #313, opened 2024-05-29
Cache `BTree` instances


kevmo314:
I take back the comment I made about instantiating BTree's freely, it's a lot more performant to maintain the cache. Let's take a slightly different approach though:

https://github.com/kevmo314/appendable/blob/main/src/db/database.ts#L208-L216

We will want to cache the relevant nodes in the BTree directly by adding two caches:

1. Add a root node cache here: https://github.com/kevmo314/appendable/blob/6c8d4eceff06ec6eeba2fc5fea54d28b74acb629/src/btree/btree.ts#L50
2. Move the child resolution to `BPTreeNode`: https://github.com/kevmo314/appendable/blob/6c8d4eceff06ec6eeba2fc5fea54d28b74acb629/src/btree/traversal.ts#L62-L64
3. Add a cache in each node to cache the n children in an array in the node object.

Lastly, we'll want to update the query search here: https://github.com/kevmo314/appendable/blob/main/src/db/database.ts#L148-L156

1. Move this out of the loop, this should be safe because all the search ops operate on the same field and thus share the same underlying BTree.

Doing those four steps should result in the queries on full text search dropping dramatically, I think it will cut our latency in half ish or a little more.

PR #316, opened 2024-05-29
feat: move child resolution to bptree node

friendlymatthew:
Step 2 in #313

Note, when we call `BPTreeNode.fromMemoryPointer()`, it creates a new node with attributes passed in.


src/btree/node.ts
```diff
@@ -56,6 +62,25 @@ export class BTreeNode {
     return this.internalPointers.length + this.leafPointers.length;
   }
 
+  async readChildNode(index: number) {
```
kevmo314, line 65:
```suggestion
  async child(index: number) {
```

This can be called just `child`, it's cleaner :)

kevmo314 (approved):

PR #319, opened 2024-05-30
feat: hnsw

friendlymatthew:
The first draft of Hnsw was messy, this PR refactors the existing code and adds tests that asserts the behavior more clearly.

`Vector` is a struct that contains a list of friends relative to that vector's position. Vector_test.go has a series of tests that checks the euclidian distance, correctly asserts the level, and links between two nodes


pkg/hnsw/vector.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+type Point []float32
+
+type Vector struct {
+	id    Id
+	point Point
+
+	friends []*BaseQueue
+}
```
kevmo314, line 15: You might be able to simplify this struct even further by instead caching the distances for each adjacency in the queue itself.

pkg/hnsw/vector.go
```diff
@@ -0,0 +1,122 @@
...
+func (v *Vector) InsertFriendsAtLevel(level int, friend *Vector) {
+	if !v.HasLevel(level) {
+		panic("failed to insert friends at level, as level is not valId")
+	}
+
+	if friend.id == v.id {
+		panic("cannot insert yourself to friends list")
+	}
+
+	dist := v.EuclidDistance(friend)
```
kevmo314, line 59: Specifically here, maybe make this insert API something like

```
func (v *Vector) Insert(level int, friend Point, dist float64)
```

This would make the distance functions more state-less.

kevmo314 (approved):

pkg/hnsw/vector.go
friendlymatthew: Done

pkg/hnsw/vector.go
friendlymatthew: > You might be able to simplify this struct even further by instead caching the distances for each adjacency in the queue itself.

Sorry, I'm a bit confused, what do you mean by caching?

kevmo314: Oh, it actually does track the distance here: https://github.com/kevmo314/appendable/blob/3aeeec2039166c2b11f59214c4e349716fef57e7/pkg/hnsw/eucqueue.go#L47

Is it necessary to have the euclidean distance functions on the vector struct in that case? It might be easier to pass in the distance as an argument.

friendlymatthew: > Is it necessary to have the euclidean distance functions on the vector struct in that case?

IMO, yes because we compute these distances on the fly relative to the vector struct's `Point`. Euclidian distance-wise, I figured it would be nice to use a `func (v *Vector)`.

kevmo314: Right, but we don't need to recompute them all the time because they're already stored in the queue? So the distance is only needed at insert time.

friendlymatthew: Oh I see. Yeah, I updated the PR so that the friend insertion has the following header:

```go
func (v *Vector) InsertFriendsAtLevel(level int, friendId Id, dist float32) {
```

pkg/hnsw/vector.go
```diff
@@ -0,0 +1,119 @@
...
+import (
+	"errors"
+	"math"
+)
+
+type Point []float32
+
+type Vector struct {
+	id    Id
+	point Point
```
kevmo314, line 12: Yeah, so this is getting closer. Note that the only user of `point` is the euclidean distance function, so if we don't store it all together then `Vector` can be

```
type Vector struct {
  friends []*BaseQueue
}
```

which is quite a bit cleaner (in particular because you could go as far as getting rid of `Vector` all together with some clever writing.

This would require pulling out euclidean distance into a pure function instead of on the struct but I think that's a good idea.

pkg/hnsw/friends.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	"errors"
+	"math"
+)
+
+type Point []float32
+
+type Friends struct {
+	friends []*BaseQueue
+}
```
friendlymatthew, line 12: Since a API is helpful during the insertion and search logic, `Friends` serves as a struct over the priority queue.

PR #322, opened 2024-06-03
feat: assert search layer

friendlymatthew:
Upon insertion, we need to first find the closest entry node.

This means searching all layers from [qTopLayer + 1..TopLayer]

### DIFF
--- a/pkg/hnsw/hnsw.go
+++ b/pkg/hnsw/hnsw.go
@@ -2,46 +2,157 @@ package hnsw
 
 import (
 	"fmt"
+	"math"
+	"math/rand"
 )
 
 type Id = uint
 
+var ErrNodeNotFound = fmt.Errorf("node not found")
+
 type Hnsw struct {
 	vectorDimensionality int
 
-	Vectors map[Id]*Friends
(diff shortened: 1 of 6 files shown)


pkg/hnsw/hnsw.go
```diff
@@ -2,46 +2,157 @@ package hnsw
...
+
+				// if distance(ccFriend, q) < distance(f, q)
+				ccFriendDistToQ := EuclidDistance(*ccFriendPoint, *q)
+				if ccFriendDistToQ < furthestFoundNN.dist || foundNNToQ.Len() < numNearestToQToReturn {
+					candidatesForQ.Insert(ccFriendId, ccFriendDistToQ)
+					foundNNToQ.Insert(ccFriendId, ccFriendDistToQ)
+
+					if foundNNToQ.Len() > numNearestToQToReturn {
+						_, err = foundNNToQ.PopItem()
+						if err != nil {
```
kevmo314, line 103:
```suggestion				
						if _, err := foundNNToQ.PopItem(); err != nil {
```

kevmo314 (approved):

PR #330, opened 2024-06-05
Knnsearch

friendlymatthew:
After inserting (0, 0), (3, 3), (4, 4), (5, 5).

When searching for the top 2 closest points to q = (5, 5), we should expect the following queue from the knnsearch:

top 2:
id: 3, point: (5, 5)
id: 2, point(4, 4)

### DIFF
--- a/pkg/hnsw/hnsw.go
+++ b/pkg/hnsw/hnsw.go
@@ -148,6 +148,20 @@ func (h *Hnsw) findCloserEntryPoint(q *Point, qFriends *Friends) *Item {
 	return epItem
 }
 
+func (h *Hnsw) selectNeighbors(nearestNeighbors *BaseQueue) (*BaseQueue, error) {
+	maxNearestNeighbors := FromBaseQueue(nearestNeighbors, MaxComparator{})
+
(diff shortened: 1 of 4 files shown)


pkg/hnsw/hnsw.go
```diff
@@ -148,18 +152,138 @@ func (h *Hnsw) findCloserEntryPoint(q *Point, qFriends *Friends) *Item {
...
+
+	for maxNearestNeighbors.Len() > h.M {
+		_, err := maxNearestNeighbors.PopItem()
+
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return FromBaseQueue(maxNearestNeighbors, MinComparator{}), nil
```
kevmo314, line 166: Is this necessary? It seems somewhat expensive to reheapify the list every iteration.

kevmo314: Yeah, reading through the code I think we can just implicitly pop the closest `h.M` items instead of reheapifying to produce a heap with the correct size. That will save two calls to `FromBaseQueue`

friendlymatthew:
Closing and will open new PR

PR #333, opened 2024-06-07
Feat: cluster c search

friendlymatthew:
Screenshot 2024-06-07 at 4 14 33 PM

Given a multi cluster graph, and a `q` = `(2, 2)`, we want to make sure the closest neighbors are pulled from both clusters and in correct order.

### DIFF
--- a/pkg/hnsw/hnsw_test.go
+++ b/pkg/hnsw/hnsw_test.go
@@ -35,11 +35,9 @@ var clusterB = []Point{
 	{4.01, 4.3},
 }
 
-func SetupClusterHnsw(cluster []Point) (*Hnsw, error) {
-	efc := 4
-
+func SetupClusterHnsw(cluster []Point, efc, maxConnections int) (*Hnsw, error) {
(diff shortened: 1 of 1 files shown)


pkg/hnsw/hnsw_test.go
```diff
@@ -559,4 +557,45 @@ func TestHnsw_KnnSearch(t *testing.T) {
 			t.Fatalf("expected closest points to be %v, got %v", expectedIds, gotIds)
 		}
 	})
+
+	t.Run("cluster c search", func(t *testing.T) {
+		clusterC := append(append([]Point{}, clusterA...), clusterB...)
+
+		clusterCLen := len(clusterC)
+
+		h := NewHnsw(2, clusterCLen, clusterCLen, Point{0, 0})
```
friendlymatthew, line 566: This means when we `Insert` every vector, the graph will consider every point as a candidate. In addition, every point's number of neighbors is `clusterCLen`.

PR #334, opened 2024-06-10
Knnsearch benchmark

friendlymatthew:
Screenshot 2024-06-10 at 10 34 37 AM

Screenshot 2024-06-10 at 10 38 20 AM


friendlymatthew:
The plot's x-axis represents the number of points generated and inserted into the hnsw before searching, the y-axis is the average time it took to search for a point in nanoseconds.

pkg/hnsw/hnsw_test.go
```diff
@@ -599,3 +600,46 @@ func TestHnsw_KnnSearch(t *testing.T) {
...
+
+	for i := 0; i < numPoints; i++ {
+		points[i] = Point{float32(i), float32(i), float32(i)}
+	}
+
+	return points
+}
+
+func BenchmarkHnsw_KnnSearch(b *testing.B) {
+	numPointsRange := []int{1, 10, 100, 1000, 10000, 100000, 1000000}
```
kevmo314, line 615: Can you run this with powers of two? Might make the trend a little clearer.

kevmo314 (commented):
Very nice! Could you try fitting an n log n curve? You can do this in python or something as a one-off, just wanna see how well it fits.

PR #337, opened 2024-06-11
assert every point's friends queue has correct max size

friendlymatthew:
Note, the level generator hallucinates if you use super small numbers that are relatively close for both `efc` and `M`.


pkg/hnsw/hnsw.go
```diff
@@ -207,30 +207,39 @@ func (h *Hnsw) InsertVector(q Point) error {
...
+			maxNumberOfNeighbors := h.M
+			if level == 0 {
+				maxNumberOfNeighbors = h.Mmax0
+			}
+
+			eConnections := neighborFriendsAtLevel
+			if eConnections.Len() > maxNumberOfNeighbors {
+				var items []*Item
+
+				for !neighborFriendsAtLevel.IsEmpty() {
```
kevmo314, line 230: This is likely going to be pretty slow. To speed it up, we probably want to use a min-max heap instead: https://en.wikipedia.org/wiki/Min-max_heap

This doesn't exist in the go std library but you can take inspiration from this library: https://github.com/esote/minmaxheap

PR #342, opened 2024-06-14
Distheap

friendlymatthew:
A minmax heap with a typespecific interface


pkg/hnsw/heap.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	return (*d).items[i], nil
+}
+func (d *DistHeap) PopMinItem() (*Item, error) {
+	if d.IsEmpty() {
+		return nil, EmptyHeapError
+	}
+
+	return Pop(d).(*Item), nil
```
kevmo314, line 79: This still seems to use the old interface, should this be `d.Pop()` instead?

friendlymatthew: Whoops. Done

pkg/hnsw/heap.go
```diff
@@ -0,0 +1,115 @@
...
+	d := &DistHeap{items: items, visited: make(map[Id]bool)}
+	Init(d)
+	return d
+}
+func (d *DistHeap) PeekMinItem() (*Item, error) {
+	if d.IsEmpty() {
+		return nil, EmptyHeapError
+	}
+
+	return (*d).items[0], nil
```
kevmo314, line 51:
```suggestion
	return d.items[0], nil
```

pkg/hnsw/heap.go
```diff
@@ -0,0 +1,115 @@
...
+	if l < n && !d.Less(l, i) {
+		i = l
+	}
+
+	r := rchild(0)
+	if r < n && !d.Less(r, i) {
+		i = r
+	}
+
+	return (*d).items[i], nil
```
kevmo314, line 72:
```suggestion
	return d.items[i], nil
```

pkg/hnsw/heap.go
```diff
@@ -0,0 +1,115 @@
...
+	PopMinItem() (*Item, error)
+	PopMaxItem() (*Item, error)
+	Insert(id Id, dist float32)
+	IsEmpty() bool
+	Len() int
+	Less(i, j int) bool
+	Swap(i, j int)
+	Push(x *Item)
+	Pop() *Item
+}
```
kevmo314, line 26: It would also be fine to not even have this interface and make the functions in `minmax.go` class functions which would remove the need for this file all together. The only reason `heap.Interface` exists is to allow external implementers but we won't have any of those.

kevmo314 (approved):

PR #343, opened 2024-06-14
feat: faster update

friendlymatthew:
`Insert` is the hottest function we call when dealing with `DistHeap`. However, `Insert` is actually `Upsert`. If the id already exists, we iterate through and call `Fix()`. Worst case this is `O(n)`. 

Now updating should be bound to `Fix()` which is `O(log n)`


kevmo314 (approved):
Nice find!

PR #344, opened 2024-06-14
remove unused methods from `DistHeap`

friendlymatthew:


pkg/hnsw/heap.go
```diff
@@ -23,24 +23,6 @@ func NewDistHeap() *DistHeap {
...
-
-	return d
-}
-
-func (d *DistHeap) Init() {
-	n := d.Len()
-	for i := n/2 - 1; i >= 0; i-- {
-		d.down(i, n)
-	}
-}
```
friendlymatthew, line 43: I can keep this if you want @kevmo314, We just never have this use case since we dynamically build our heaps from scratch

kevmo314: Nah, we can always readd it later :)

PR #349, opened 2024-06-20
Convert to array-based points map

friendlymatthew:


pkg/hnsw/hnsw.go
```diff
@@ -194,7 +191,7 @@ func (h *Hnsw) InsertVector(q Point) error {
 
 	qFriends := NewFriends(qTopLevel)
 	h.friends[qId] = qFriends
-	h.points[qId] = &q
+	h.points = append(h.points, &q)
```
friendlymatthew, line 194: note: this makes it sequential

kevmo314 (approved):

PR #356, opened 2024-06-24
Btree insert

friendlymatthew:
I'm following: 
```
Insertion 
1) Initialize x as root. 
2) While x is not leaf, do following 
..a) Find the child of x that is going to be traversed next. Let the child be y. 
..b) If y is not full, change x to point to y. 
..c) If y is full, split it and change x to point to one of the two parts of y. If k is smaller than mid key in y, then set x as the first part of y. Else second part of y. When we split y, we move a key from y to its parent x. 
3) The loop in step 2 stops when x is leaf. x must have space for 1 extra key as we have been splitting all nodes in advance. So simply insert k to x. 

Note that the algorithm follows the Cormen book. It is actually a proactive insertion algorithm where before going down to a node, we split it if it is full. The advantage of splitting before is, we never traverse a node twice. If we don’t split a node before going down to it and split it only if a new key is inserted (reactive), we may end up traversing all nodes again from leaf to root. This happens in cases when all nodes on the path from the root to leaf are full. So when we come to the leaf node, we split it and move a key up. Moving a key up will cause a split in parent node (because the parent was already full). This cascading effect never happens in this proactive insertion algorithm. There is a disadvantage of this proactive insertion though, we may do unnecessary splits. 
```

However, I'm a bit confused on step 2, specifically what to do with the rightChild.

2) While x is not leaf, do following 
..a) Find the child of x that is going to be traversed next. Let the child be y. 
..b) If y is not full, change x to point to y. 
..c) If y is full, split it and change x to point to one of the two parts of y. If k is smaller than mid key in y, then set x as the first part of y. Else second part of y. When we split y, we move a key from y to its parent x. 

after we split the child, we have `leftNode` and `rightNode` , where `leftNode` is just the updated child. But what do we do with the rightNode? We add the midkey into the parent node as specified.

### DIFF
--- a/pkg/bptree/bptree.go
+++ b/pkg/bptree/bptree.go
@@ -42,12 +42,12 @@ type TraversalRecord struct {
 
 type TraversalIterator struct {
 	tree    *BPTree
-	key     ReferencedValue
+	key     pointer.ReferencedValue
 	records []TraversalRecord
 	err     error
 }
 
-func (p *TraversalIterator) Key() ReferencedValue {
+func (p *TraversalIterator) Key() pointer.ReferencedValue {
 	return p.records[0].node.Keys[p.records[0].index]
 }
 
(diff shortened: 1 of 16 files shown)


kevmo314:
> But what do we do with the rightNode? We add the midkey into the parent node as specified.

Right node is added to the parent directly right of the left node. See: https://www.youtube.com/watch?v=xEOdJd7GcmQ

This is probably not intuitive because you're used to binary trees. The parent can hold n+1 nodes so both the left and right node are attached.

friendlymatthew:
> > But what do we do with the rightNode? We add the midkey into the parent node as specified.
> 
> Right node is added to the parent directly right of the left node. See: https://www.youtube.com/watch?v=xEOdJd7GcmQ
> 
> This is probably not intuitive because you're used to binary trees. The parent can hold n+1 nodes so both the left and right node are attached.

Gotcha. So this would mean creating a new page for `rightNode`. as well as updating the parent's pointer with the right child offset

kevmo314:
> > > But what do we do with the rightNode? We add the midkey into the parent node as specified.
> > 
> > 
> > Right node is added to the parent directly right of the left node. See: [youtube.com/watch?v=xEOdJd7GcmQ](https://www.youtube.com/watch?v=xEOdJd7GcmQ)
> > This is probably not intuitive because you're used to binary trees. The parent can hold n+1 nodes so both the left and right node are attached.
> 
> Gotcha. So this would mean creating a new page for `rightNode`. as well as updating the parent's pointer with the right child offset

Yep, correct.

PR #358, opened 2024-06-26
`btree` insert root

friendlymatthew:


friendlymatthew:
@kevmo314 inserting the root element seems to work great!

PR #362, opened 2024-06-27
feat: marshal, unmarshal, size

friendlymatthew:


kevmo314 (approved):

friendlymatthew:
@kevmo314 -- just want to note that we cast every element (`float32`) in `Vector` as a `uint32`

kevmo314:
> @kevmo314 -- just want to note that we cast every element (`float32`) in `Vector` as a `uint32`

Yeah that's fine :)

PR #364, opened 2024-07-01
Btree random insertion test

friendlymatthew:


pkg/btree/btree.go
```diff
@@ -48,3 +51,173 @@ func (t *BTree) readNode(offset uint64) (*BTreeNode, error) {
...
+		parent.Ids = append(parent.Ids[:index+1], parent.Ids[index:]...)
+		parent.Ids[index] = key
+		parent.Vectors = append(parent.Vectors[:index+1], parent.Vectors[index:]...)
+		parent.Vectors[index] = vector
+	}
+
+	// this should be the leaf node
+	if int(parent.Size()) > t.PageFile.PageSize() {
+		panic("too fat!")
+	}
```
friendlymatthew, line 181: When we reach the point where we finally serialize the `leaf` node post top-down traversal, that `leaf` size can exceed a page. I was going to split, but is creating an intermediate node with only one element (the mid key) be sufficient?

kevmo314: Hmm I think splitting while traversing downwards won't work because of this problem. Adding an intermediate node isn't great because there is a degenerate sequence of inserts that will always trigger an intermediate node insertion leading to bad asymptotic performance.

Specifically, right now you're splitting the tree as you're traversing down to the leaf node but instead it might be easier to insert it first and then propagate the split back _up_ the tree. This would fix the issue as you wouldn't need to create an intermediate node.

friendlymatthew: Sounds good

PR #365, opened 2024-07-01
add guard to `Size`

friendlymatthew:
A common trap when creating a `BTree` is to zero out `VectorDim`. However, a `VectorDim` of 0 proceeds to screw up the `Size` calculation.


friendlymatthew:
Note: Instead, it'll panic.

PR #369, opened 2024-07-08
feat: check in max level for every friend id

friendlymatthew:
Per the discord comments about flushing `Friends` to disk.

DMR till #368

### DIFF
--- a/pkg/hnsw/friends.go
+++ b/pkg/hnsw/friends.go
@@ -8,7 +8,8 @@ import (
 type Point []float32
 
 type Friends struct {
-	friends []*DistHeap
+	friends   []*DistHeap
+	maxLevels map[Id]int
 }
 
 // NewFriends creates a new vector, note the max level is inclusive.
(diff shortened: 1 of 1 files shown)


kevmo314 (approved):

pkg/hnsw/friends.go
```diff
@@ -8,7 +8,8 @@ import (
 type Point []float32
 
 type Friends struct {
-	friends []*DistHeap
+	friends   []*DistHeap
+	maxLevels map[Id]int
```
kevmo314, line 12: Actually would this work as a slice since `Id` is sequential?

friendlymatthew:
@kevmo314 I can't directly respond to your comment, but since the `Hnsw` itself takes in a mapping from every node to it's unique `Friends` data structure.

Since the entry point is dynamically set, I'm not sure if we can ensure that the node-specific `Friends` datastructure will store `Ids` sequentially.

PR #370, opened 2024-07-09
feat: flush `Friends` to disk

friendlymatthew:


pkg/hnsw/friends.go
```diff
@@ -63,6 +65,45 @@ func (v *Friends) GetFriendsAtLevel(level int) (*DistHeap, error) {
...
+
+	for i := 0; i < numNeighbors; i++ {
+		if copyLevel0.IsEmpty() {
+			// write out max values here
+			continue
+		}
+
+		closestItem, err := copyLevel0.PopMinItem()
+		if err != nil {
+			panic(fmt.Sprintf("failed to find closest item in friends: %v", err))
```
kevmo314, line 90: Return the err instead of panicking, it tends to be pretty annoying to work around this later.

kevmo314 (approved):

PR #371, opened 2024-07-10
feat: adjacency manager

friendlymatthew:
This PR creates a `VectorPageManager` struct. When adding a node, the `VectorPageManager` is responsible for slotting `point id` in the btree and the respective closest friends in the bptree.

### DIFF
--- a/pkg/vectorpage/manager.go
+++ b/pkg/vectorpage/manager.go
@@ -1,0 +1,53 @@
+package vectorpage
+
+import (
+	"fmt"
+	"github.com/kevmo314/appendable/pkg/bptree"
+	"github.com/kevmo314/appendable/pkg/btree"
+	"github.com/kevmo314/appendable/pkg/hnsw"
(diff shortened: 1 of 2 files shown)


pkg/vectorpage/manager.go
```diff
@@ -0,0 +1,53 @@
...
+
+type VectorPageManager struct {
+	btree   *btree.BTree
+	vectors []*hnsw.Point
+
+	bptree       *bptree.BPTree
+	neighborhood map[hnsw.Id]*hnsw.Friends
+}
+
+func NewVectorPageManager(btree *btree.BTree, bptree *bptree.BPTree, vectors []*hnsw.Point, neighborhood map[hnsw.Id]*hnsw.Friends) *VectorPageManager {
```
kevmo314, line 21: Does this also need the backing pagefile?

kevmo314 (approved):

PR #375, opened 2024-07-19
feat: refactor vector manager

friendlymatthew:
Follow up from yesterday's 1-1:

The `VectorPageManager` should encapsulate the `hnsw`. As we iterate by table row, the `VectorPageManager` is responsible for updating the `hnsw`, then serializing the node state to disk.

Note this is a highly stateful action as we don't know a node's `closest` friends until all other nodes have been `considered`.

### DIFF
--- a/pkg/vectorpage/manager.go
+++ b/pkg/vectorpage/manager.go
@@ -11,36 +11,44 @@ import (
 type HNSWAdjacencyPage [16][8]uint32
 
 type VectorPageManager struct {
-	btree   *btree.BTree
-	vectors []*hnsw.Point
+	btree *btree.BTree
+	// vectors []*hnsw.Point
 
-	bptree       *bptree.BPTree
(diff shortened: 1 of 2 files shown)


pkg/vectorpage/manager.go
```diff
@@ -11,36 +11,44 @@ import (
...
 type VectorPageManager struct {
-	btree   *btree.BTree
-	vectors []*hnsw.Point
+	btree *btree.BTree
+	// vectors []*hnsw.Point
 
-	bptree       *bptree.BPTree
-	neighborhood map[hnsw.Id]*hnsw.Friends
+	bptree *bptree.BPTree
+	// neighborhood map[hnsw.Id]*hnsw.Friends
```
friendlymatthew, line 18: I can remove the comments, just wanted to show that we're fully encapsulating the `hnsw` inside the `VectorPageManager`

